PR #1, opened 2016-09-23
BEANUTILS-497: Fix javadoc doclint errors

bond-:
Finally after a ton of changes the build is successful. I double checked all my changes in Javadoc but it would be good if someone would also give a +1.

Build command: ```mvn clean javadoc:javadoc -DadditionalJOption=-Xdoclint:all```


stain:
Thanks alot! I'll use this in the site for the 1.9.3 release. I've added you as a contributor in the `pom.xml`, hope that's OK.

PR #8, opened 2019-10-19
Convert Collections4 to java.util.function

melloware:
@garydgregory I think this is what you were thinking for Predicate?

Feel free to let me know if have the generics wrong or anything else you want changed.
I kept the evaluate() method as is even though it just delegates to test() method.  Wasn't sure if that was what you wanted either?

**Notes:**
Should evaluate method be marked Deprecated since test() is the real function in java.util.Predicate?

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/BeanPredicate.java
+++ b/src/main/java/org/apache/commons/beanutils2/BeanPredicate.java
@@ -18,8 +18,8 @@
 package org.apache.commons.beanutils2;
 
 import java.lang.reflect.InvocationTargetException;
+import java.util.function.Predicate;
 
-import org.apache.commons.collections4.Predicate;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/beanutils2/BeanPredicate.java
```diff
@@ -50,6 +50,18 @@ public BeanPredicate(final String propertyName, final Predicate predicate) {
...
+    
+    /**
+     * Evaluates the given object by applying the {@link #getPredicate()}
+     * to a property value named by {@link #getPropertyName()}.
+     *
+     * @param object The object being evaluated
+     * @return the result of the predicate evaluation
+     * @throws IllegalArgumentException when the property cannot be evaluated
+     */
+    public boolean evaluate(Object object) {
```
jvz: You could probably mark this `@Deprecated`.

garydgregory: No need to deprecated, we can just remove it since we are in a major release.

PR #10, opened 2019-10-21
BEANUTILS-528: New converters for UUID, URI, and Path.

melloware:


src/main/java/org/apache/commons/beanutils2/converters/PathConverter.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param type Data type to which this value should be converted.
+     * @param value The input value to be converted.
+     * @return The converted value.
+     * @throws Throwable if an error occurs converting to the specified type
+     * @since 2.0
+     */
+    @Override
+    protected <T> T convertToType(final Class<T> type, final Object value) throws Throwable {
+        if (Path.class.equals(type)) {
+            return type.cast(Paths.get(String.valueOf(value)));
```
jvz: Doing a cast here is redundant unless you change the `if` statement to check `Path.class.isAssignableFrom(type)`. Then you can support `Path` and subclasses.

garydgregory: Doesn't the same apply to the URI converter since it is not a final class?

melloware: URI is a final class though right?  I modeled it after the URLConverter that already exists and both look like final classes?

garydgregory: URI is _not_ final.

melloware: Maybe we are talking about 2 different things and I am misunderstanding?  I am looking at JDK8 source code and URI is declared...
```java
public final class URI extends Object implements Comparable<URI>, Serializable
```
https://docs.oracle.com/javase/8/docs/api/java/net/URI.html

Are we talking about different things?

garydgregory: Yes, my bad, I was mistakenly looking at `com.sun.org.apache.xerces.internal.util.URI`.

PR #12, opened 2019-10-22
Remove BeanUtils2/ConvertUtils2

melloware:
@garydgregory I think this is what you were asking for on the mailing list.  Make ConvertUtils2/BeanUtils2 the default behavior by moving that code into the main Convert/Bean Utils and removing the old classes?


garydgregory:
Let's wait to see if Nial see the message on the dev ML.

melloware:
@garydgregory Any thoughts on this since we haven't heard from Nial?

melloware:
closing PR as I rebased a better one.

PR #17, opened 2019-10-23
[BEANUTILS-468] MappedPropertyDescriptor should override hashCode /equals.

melloware:


garydgregory:
This does not look right since `mappedPropertyTypeRef` is not taken into account and two instances can be equal even when instance variables in super classes are different.

melloware:
I was hoping you would notice that.  When I originally included it Eclipse warned "The field type java.lang.Reference does not redeclare equals() and hashcode() - the resulting code may not work correctly".

I can add it in I just was a little worried about the warning above?

melloware:
I tried it out and I fix it but the unit test is now pretty useless because I can never get the same euqals/hashcode for two identically created values.

PR #24, opened 2020-05-26
fix travis-ci scripts

XenoAmess:


XenoAmess:
test on jdk14 fails.
I just moved jdk14 to allow failure.
Once you fixed it, you can remove it from allow failure.

melloware:
Wouldn't it be better to help fix the tests rather than temporarily ignore failures?

garydgregory:
> Wouldn't it be better to help fix the tests rather than temporarily ignore failures?

+1

Or at least understand and document the failure, possibly disabling the test under certain condition with a JUnit `Assume` call.

XenoAmess:
> Wouldn't it be better to help fix the tests rather than temporarily ignore failures?

1. yes it is.
2. I have no ideas about how to fix these bugs.
3. I do think the bugs I found be minor problems, as this repo can work for such a long time without handling(or even noticing) the bug I mentioned.
4. I assume there must be somebody who knows more than me about how to solve these bugs, for example some guys who knows more about this repo/this test class.

melloware:
@XenoAmess You can kill this PR we have the build working properly now on JDK15+

garydgregory:
Close per previous comment.

PR #38, opened 2020-08-28
Spotless Formatter and License Header Check

melloware:
Eclipse formatter which follows all the CheckStyle and Apache Commons guidelines for formatting. 4 Spaces, 120 line limit, proper whitespace handling etc.

Window->Preferences->Java->Code Style->Formatter and then press the Import button.

Point to eclipse-apache-formatter.xml.

You can format the entire project in one shot by right clicking on the package org.apache.commons.beanutils2 and selecting Source->Format.

OR...

Run `mvn spotless:check` or `mvn spotless:apply` to either format code or validate the format.

@aherbert I turned OFF all Javadoc formatting and now this looks much better.  its not perfect but I don't know if there is a perfect formatter out there.   This is to get the project in a good starting state for 2.0.0 release.


pom.xml
```diff
@@ -16,417 +16,428 @@
...
-  <parent>
-    <groupId>org.apache.commons</groupId>
-    <artifactId>commons-parent</artifactId>
-    <version>52</version>
-  </parent>
-  <modelVersion>4.0.0</modelVersion>
-  <artifactId>commons-beanutils2</artifactId>
-  <version>2.0.0-SNAPSHOT</version>
-  <name>Apache Commons BeanUtils</name>
+    <parent>
```
melloware, line 19: I formatted the pom.xml because it had a mix of tabs and spaces so I standardized it to all SPACES.

melloware:
Now I am able to run both `mvn formatter:format` and 'mvn checkstyle:check` and both work correctly.

rmannibucau:
Maybe use spotless for eclipse formatter cause it uses m2 dependencies and not p2 as formatter plugin which keeps checking for updates - since p2 are mutable even for releases. Will avoid a lot of latency with this.

melloware:
@rmannibucau  By Spotless I a assume you mean this? https://awesomeopensource.com/project/diffplug/spotless

As for out plugin it s not run as part of the build process we have CheckStyle for that.  This simply is run on `mvn formatter:format` to get the code in a good state following Apache Guidelines.

rmannibucau:
Yes
Spotless is also used by the asf and avoids issues, ran by default or not so think it is saner.

melloware:
Do you have an example Apache project I can look at to model it after?

pom.xml
XenoAmess: > I formatted the pom.xml because it had a mix of tabs and spaces so I standardized it to all SPACES.

glad to see any tab deleted.
I HATE TABS.

melloware:
Switched to the Spotless plugin and its great.  Also added a license header check whcih comes as part of the spotless plugin.

aherbert:
Previously this PR had only the configuration changes. It now has the result of the formatter being applied to the entire project. Can you undo the bulk changes to just leave the formatter plugin configuration. This makes it easy to apply locally and diff all the changes that are created by the formatter.

melloware:
@aherbert submitted separate PR: https://github.com/apache/commons-beanutils/pull/39

Will close this one if the other is accepted.

melloware:
Sorry meant to say we should merge this one after the other once since I had to manually fix something things to get both Spotless and CheckStyle to both pass every time.

PR #47, opened 2020-11-03
Add Converters for Additional Types

SethFalco:
This adds the following converters, with test cases:
* Color 
* Dimension
* InetAddress
* Locale
* Pattern
* Point

You can see the test cases for example inputs/outputs.

I've just taken my PR from DeltaSpike (https://github.com/apache/deltaspike/pull/109) and converted the converters that didn't exist here to work with this project instead. This is just me acting on a comment in the PR (https://github.com/apache/deltaspike/pull/109#issuecomment-720469005) in case it does happen, I figured I could have the PR here already.

This does not port all changes from my other PR, for example I implemented some existing converters differently, but I'll PR those separately, so they can just be closed if they're unfavorable.

PR #49, opened 2020-11-04
Updated Existing Converters

SethiPandi:
This updates some existing converters to the implementations referenced in https://github.com/apache/commons-beanutils/pull/47.
I didn't bother moving code over if the only differences were [conditional checks](https://github.com/apache/deltaspike/pull/109/files#diff-63785c6a056b9caf8429e7ad1aa80c1d7f3ef308b0a7575a0668ae7f00742a2f), or [constructor parameters](https://github.com/apache/deltaspike/pull/109/files#diff-c915606a7619184f993de23365598c0305831363c9a53ec92129ef266af32631).

All existing test cases still pass, and new ones have been added.

**Character**
Accepts hex input, to specify a character via its character code instead.

Intended where there are concerns regarding the environment, such as system/file encoding between clients, applications, and servers. (I've worked somewhere before where we couldn't put £ or € in the source. ^-^')

Original plan was to enforce single character Strings, or a hexadecimal integer only. Since a CharacterConverter already exists in this repo which allows multiple character Strings, I've kept that around but added the hexadecimal support on top. That means `0xA3` would return `£`, not `0`. Not sure if this is a good idea though as it may cause confusion along with allowing multiple character Strings.

**Duration**
Will now accept number of milliseconds if it can't parse the input as a Duration string.
Intended to just be an easier way to define it rather than needing to know the Duration syntax.

**Enum**
Will now accept types to the normal `Enum` class, for example if assigning to `Enum`, it can accept input from `java.time.DayOfWeek#MONDAY` or `java.time.DayOfWeek.MONDAY` to get the constant `MONDAY` from the `DayOfWeek` enum type.

Intended for cases where an enum is required, but not a specific type, or multiple types. My use case is having certain enums (such as states, actions, or categories of events) which are localized, in the language file; the key being full enum name which can be converted to an EnumMap at startup. (Might be a dumb idea, haven't really tested this yet... but implemented this part anyway.)

**Instant**
Can now convert Instants (another DateTime type).

I wanted this to support numeric Strings (ie milliseconds as a String) as well but I didn't do this now. I'll explore the abstract DateTimeConverter more later. 

**Period**
Will now accept a number of days if it can't parse the input as a Period string.
Intended for the same reason as Duration.

---

* I've updated test classes for the converters that this affects to JUnit 4. (ie `@Test` annotations, and no longer extending `TestCase`)
* In tests I've updated the use of some deprecated methods to what is recommended instead. i.e. (`new Character()` → `Character.valueOf()`) (`new Character()` is deprecated from Java 9, but `Character.valueOf()` is already available, so figured I could just move it already.)


melloware (approved):
Really good improvements IMHO.

melloware:
Looks like your EnumConverter test is failing the build?

SethiPandi:
> Looks like your EnumConverter test is failing the build?

Ahh shoot, I'll fix that then. :thinking: 
Not sure but could be 'cause I have JDK 11 as it seems to build/test for me locally. I'll swap out and see if it makes a difference.

--- 

I actually just noticed a small differences between the old and new EnumConverter, though.
The original was case-insensitive, but the new one is case-sensitive. 

I was going to change it back, but verify something first as that might not be ideal:

```java
public enum SpecialEnum {
    HELLO,
    hello
}
```

Despite the fact it breaches naming conventions, it's technically still possible to have multiple enum constants with the same name if we ignore case, so I think it's better to be case-sensitive to avoid choosing the wrong one.

I think changing it to case-sensitive might be considered a breaking change, though?
Do you have any thoughts on if I should leave it case-sensitive, or revert it to be non-case sensitive?

melloware:
I think if you look a the history Enum is a new converter for 2.0 which has not been released and never existing in 1.9.X branch so you should be OK and it shouldn't be a breaking change.

SethiPandi:
Not 100% sure why, but seems on Java 8 I need to add a cast to Enum. :thinking:

melloware (approved):

melloware:

SethFalco:
I've rebased the PR with `master` now, it should be up to date including the JaCoCo 0.8.6 commit.

src/test/java/org/apache/commons/beanutils2/converters/PeriodConverterTestCase.java
```diff
@@ -86,31 +59,50 @@ public void testSimpleConversion() throws Exception {
...
-            fail("Unsupported type could be converted!");
-        } catch (final ConversionException cex) {
-            // expected result
-        }
+        converter.convert(Integer.class, "http://www.apache.org");
+    }
+
+    @Test
+    public void testConvertingDefault() {
+        final Period expected = Period.ofDays(2);
```
garydgregory, line 87: Both of these locals can be inlined.

src/main/java/org/apache/commons/beanutils2/converters/DurationConverter.java
```diff
@@ -74,10 +84,21 @@ public DurationConverter(final Object defaultValue) {
...
         if (Duration.class.equals(type)) {
-            return type.cast(Duration.parse((String.valueOf(value))));
+            final String stringValue = toString(value);
+
+            try {
+                return type.cast(Duration.parse(stringValue));
+            } catch (DateTimeParseException ex) {
+                try {
+                    long number = Long.parseLong(stringValue);
+                    return type.cast(Duration.of(number, ChronoUnit.MILLIS));
```
garydgregory, line 94: **-1**: This feels like a recipe for problems and confusion, the code expects _some_ number in _some_ format in _some_ scale. This is just not a Duration which is a well defined concept.

src/main/java/org/apache/commons/beanutils2/converters/PeriodConverter.java
```diff
@@ -74,7 +87,19 @@ public PeriodConverter(final Object defaultValue) {
...
     protected <T> T convertToType(final Class<T> type, final Object value) throws Throwable {
         if (Period.class.equals(type)) {
-            return type.cast(Period.parse((String.valueOf(value))));
+            final String stringValue = toString(value);
+
+            try {
+                return type.cast(Period.parse(stringValue));
+            } catch (DateTimeParseException ex) {
+                try {
+                    int number = Integer.parseInt(stringValue);
```
garydgregory, line 96: Upon revisitation **-1**: Same comment as for Duration.

src/test/java/org/apache/commons/beanutils2/converters/CharacterConverterTestCase.java
```diff
@@ -18,109 +18,97 @@
...
+        assertEquals("Default Test", Character.valueOf('C'), converter.convert(Character.class, null));
+    }
+
+    /**
+     * If a hexadecimal value is provided, we'll convert it to a
+     * a character if possible.
+     */
+    @Test
+    public void testConvertHexCharacter() {
+        final char expected = 'A';
```
garydgregory, line 101: Inline local vars.

src/test/java/org/apache/commons/beanutils2/converters/CharacterConverterTestCase.java
```diff
@@ -18,109 +18,97 @@
...
+        final char actual = converter.convert(Character.class, "0x41");
+
+        Assert.assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testConvertHexSymbol() {
+        final char expected = '£';
```
garydgregory, line 109: Inline local vars.

src/test/java/org/apache/commons/beanutils2/converters/ConverterTestSuite.java
```diff
@@ -41,7 +41,6 @@ public static Test suite() {
         testSuite.addTestSuite(BooleanConverterTestCase.class);
         testSuite.addTestSuite(ByteConverterTestCase.class);
         testSuite.addTestSuite(CalendarConverterTestCase.class);
-        testSuite.addTestSuite(CharacterConverterTestCase.class);
         testSuite.addTestSuite(ClassConverterTestCase.class);
```
garydgregory, line 44: Not sure why you deleted this class from this list. What am I missing? Please explain. The class is gone from git master anyway but I am still curious. Please rebase :-)

garydgregory (changes requested):
@SethFalco 
Thank you for your PR. This PR is almost to big. Please see my comments.

garydgregory (changes requested):
@SethFalco 
Thank you for your PR. This PR is almost too big. Please see my comments. It might be simpler, depending on your reply, of splitting up this PR into smaller ones which would expedite non-controversial changes like the addition of _new_ converters vs. changing existing ones.
G.

src/test/java/org/apache/commons/beanutils2/converters/ConverterTestSuite.java
SethFalco: While doing this, I migrated tests of classes I was impacting to JUnit 4. When I changed the `CharacterConverter` I noticed that the line wasn't doing anything, so I removed it assuming it was just some pre-JUnit 4 boilerplate that wasn't needed anymore. ^-^' (Only ever used JUnit 4/5.)

mohanaraosv:
@garydgregory  Some of your comments related to code format could be resolved by defining IDE specific code formatter.  WDYT.

melloware:
Yes that is why I have added a PR here to implement Spotless plugin so the code stays formatted to project standards no matter what IDE is used!

SethFalco:
> @garydgregory Some of your comments related to code format could be resolved by defining IDE specific code formatter. WDYT.

What @melloware suggested is much more appealing in general, coming from an IntelliJ user who is aware some other contributors are using Eclipse. It'd be better as a Maven/Gradle build task, Travis/GitHub action, or anything that does not bind the contributor to a particular IDE.

garydgregory:
> > @garydgregory Some of your comments related to code format could be resolved by defining IDE specific code formatter. WDYT.
> 
> What @melloware suggested is much more appealing in general, coming from an IntelliJ user who is aware some other contributors are using Eclipse. It'd be better as a Maven/Gradle build task, Travis/GitHub action, or anything that does not bind the contributor to a particular IDE.

I get it but we would have to provide a consistent approach for all components in Apache Commons IMO. We could start with just one. Right now we use Checkstyle, so we would have to convert each component's checkstyle configuration to a new system. Yes, ideally we could have one rule set for all of Commons but we don't beyond use spaces not tabs.

Anyway, please feel free to bring this up on the dev mailing list.

SethFalco:
* I've rebased with master.
* Squashed my commits.
* I believe I've resolved all of your feedback.

High hopes everything is cool, but I'll be available if you have any further feedback.

I opted to stick to updating this PR for now rather than split it since it's only getting smaller anyway. (let me know if that's a bother!)

I left some changes in converters I reverted, for example:
* updating test cases
* applying conventions to nearby code to what I was modifying

src/test/java/org/apache/commons/beanutils2/converters/CharacterConverterTestCase.java
```diff
@@ -16,112 +16,107 @@
...
-
-    /**
-     * Create Test Suite
-     *
-     * @return test suite
-     */
-    public static TestSuite suite() {
-        return new TestSuite(CharacterConverterTestCase.class);
-    }
+class CharacterConverterTestCase {
```
garydgregory, line 31: @SethFalco 
Our convention is to use public classes for test cases. Please do not change this to keep the PR focused and minimize PR noise and size.

src/main/java/org/apache/commons/beanutils2/converters/CharacterConverter.java
```diff
@@ -22,11 +22,19 @@
  * <p>
  * Can be configured to either return a <i>default value</i> or throw a
  * {@code ConversionException} if a conversion error occurs.
+ * </p>
+ *
```
garydgregory, line 26: Remove extra Javadoc blank line.

src/main/java/org/apache/commons/beanutils2/converters/InstantConverter.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.beanutils2.converters;
+
+import java.time.Instant;
+
+/**
+ * {@link org.apache.commons.beanutils2.Converter} implementation that handles conversion
+ * to and from {@link Instant} objects.
+ *
```
garydgregory, line 24: Remove extra Javadoc blank line.

src/test/java/org/apache/commons/beanutils2/converters/CharacterConverterTestCase.java
```diff
@@ -16,112 +16,107 @@
...
+        assertThrows(ConversionException.class, () -> converter.convert(Character.class, null));
     }
 
     /**
      * Test Conversion to String
      */
+    @Test
     @SuppressWarnings("unchecked") // testing raw conversion
-    public void testConvertToString() {
+    void testConvertToString() {
```
garydgregory, line 76: Don't change the public test convention we use in tests.

src/test/java/org/apache/commons/beanutils2/converters/EnumConverterTestCase.java
```diff
@@ -50,41 +48,74 @@ protected Converter<Enum<PizzaStatus>> makeConverter() {
...
+    @Test
+    void testUnsupportedType() {
+        assertThrows(ConversionException.class, () -> converter.convert(Integer.class, "http://www.apache.org"));
+    }
+
+    @Test
+    void testConvertTimeUnit() {
+        final TimeUnit expected = TimeUnit.NANOSECONDS;
+        final Enum actual = converter.convert(Enum.class, "java.util.concurrent.TimeUnit.NANOSECONDS");
+
```
garydgregory, line 90: Remove extra whitespace.

src/test/java/org/apache/commons/beanutils2/converters/EnumConverterTestCase.java
```diff
@@ -50,41 +48,74 @@ protected Converter<Enum<PizzaStatus>> makeConverter() {
...
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void testConvertDayOfWeek() {
+        final DayOfWeek expected = DayOfWeek.MONDAY;
+        final DayOfWeek actual = converter.convert(DayOfWeek.class, "java.time.DayOfWeek#MONDAY");
+
```
garydgregory, line 98: Remove extra whitespace.

src/test/java/org/apache/commons/beanutils2/converters/EnumConverterTestCase.java
```diff
@@ -50,41 +48,74 @@ protected Converter<Enum<PizzaStatus>> makeConverter() {
...
      */
-    public void testUnsupportedType() {
-        try {
-            converter.convert(Integer.class, "http://www.apache.org");
-            fail("Unsupported type could be converted!");
-        } catch (final ConversionException cex) {
-            // expected result
-        }
+    @Test
+    void testUnsupportedType() {
```
garydgregory, line 82: All tests are public (our convention).

garydgregory (changes requested):
Hi @SethFalco 
Thank you for your update. Please see minor scattered comments.

src/test/java/org/apache/commons/beanutils2/converters/CharacterConverterTestCase.java
SethFalco: Happy to do this, but just want to clarify something, as I made this change based on the JUnit User Guide:

> _Class and method visibility_
> 
> Test classes, test methods, and lifecycle methods are not required to be `public`, but they must not be `private`.
> 
> It is generally recommended to omit the `public` modifier for test classes, test methods, and lifecycle methods unless there is a technical reason for doing so – for example, when a test class is extended by a test class in another package. Another technical reason for making classes and methods `public` is to simplify testing on the module path when using the Java Module System.
>
> — https://junit.org/junit5/docs/current/user-guide/

I'll assume you'd prefer to have your convention applied for now. However, if sharing this convinces you otherwise, let me know and I'd be happy to remove the scope again. 👍🏽

SethFalco:
Thanks for the review. I've rebased and made all requested changes, including the visibility one, despite not explicitly resolving the conversation above.

SethFalco:
I've just rebased with `master`, but have also made a few changes this time:

1. I removed my modifications to `CharacterConverter` which added support for hex strings via a `0x` prefix. I can add this back in upon request. In 2025, I don't see any reason to bundle this logic when file/string encoding should just be configured properly for projects/infra. Retrospecting on the past, even in the instance I referred to in the original issue where symbols like `£` would break, we could've easily changed the file encoding to UTF-8, I just wasn't aware of why certain characters were breaking at the time.
2. My PR no longer migrates test cases to JUnit 5 as that has already been done in `master`. (I did however keep the changes that that made `CharacterConverterTest` use the same variable between test cases.)

That means the scope of the PR is:

* Add support for the `Instant` type.
* Add an explicit error for empty strings for `CharacterConverter`.
* Migrate some instances of `String.valueOf` to the `toString(value)` helper.
* Expand on the logic for the `EnumConverter` to support fully qualified Enum constants.

SethFalco:
@garydgregory Sorry to ping you, but just thought it might be the the way to go as this did not receive a review last time when I only commented.

No pressure if you're busy ofc. :+1:

SethFalco:
@garydgregory 

Rebased! I've also made the following changes:

* I've removed the `public` keyword from test classes and test functions I'd added, based on https://github.com/apache/commons-beanutils/commit/c00f87291b763283cd462a1a3ab5e562f5a3c39f and https://github.com/apache/commons-beanutils/commit/f89bdd34c43ae18f0e34189b08aa447afc9044c7.
* I replaced the regular expression in `EnumConverter` with String parsing. Back when I originally wrote this, I was overzealous about regular expression, and used it as my default choice of tool. However, I've since become more conscious of performance and appreciate a few extra lines of String parsing is often better than pattern matching, plus many don't consider them maintainable.

> ```
> Benchmark                                 Mode  Cnt     Score   Error  Units
> EnumStringToConstBenchmark.stringParsing  avgt       7122.947          ns/op
> EnumStringToConstBenchmark.usingRegex     avgt       9006.940          ns/op
> ```
> 
> — [SethFalco/java-microbenchmarks#!4](https://gitlab.com/SethFalco/java-microbenchmarks/-/merge_requests/4) _(lower is better)_

The change in implementation shouldn't affect the behavior. All test cases have identical results to before, and in my benchmark I checked a few more cases too. Let me know if you'd like me to reverse that change or have other feedback!

garydgregory:
@SethFalco 
TY for your patience, merged.

SethFalco:
> Do you think anything else is missing for `java.time`?

Hmm, I don't do Java as much as I used to, so I'm not up-to-date if there's been anything new since Java 11. (I'd only recently picked up some of my Java projects again, and upgraded them to Java 17, so I'm pretty behind too.)

The types currently supported seem to cover all of the main chrono types, and certainly all the ones I use in my projects, so I think it's cool. Ofc you can expect another PR though if the status changes!

Thank you for reviewing/merging my PR!

PR #50, opened 2020-11-04
Updated URLConverter

SethFalco:
While working on another PR, I noticed some unit tests started locking up when I lost my internet connection temporarily.
When looking into why, I learned something interesting about the `URL#equals` method.

```java
/**
 * ...
 * Two URL objects are equal if they have the same protocol, reference
 * equivalent hosts, have the same port number on the host, and the same
 * file and fragment of the file.<p>
 *
 * Two hosts are considered equivalent if both host names can be resolved
 * into the same IP addresses; else if either host name can't be
 * resolved, the host names must be equal without regard to case; or both
 * host names equal to null.<p>
 *
 * Since hosts comparison requires name resolution, this operation is a
 * blocking operation. <p>
 * ...
 */
```

In other words, the tests for the URLConverter connect to the internet and will stall the tests if it's unable to do so as it's a blocking operation.

I'm not sure what the guidelines for the repository is, but I've always been pretty firm on, configuration/environment should not affect automated tests, so I think this should be changed to no longer block when there's no internet.

To do an offline comparison, one should use `URI#equals` or `String#equals`.
I'm not sure if the fix is perfect from a QA standpoint, but it avoids this by comparing `String`s instead of `URL`s.

---

While I was at it, I also updated the URLConverter to use JUnit 4.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/converters/ConverterTestSuite.java
+++ b/src/test/java/org/apache/commons/beanutils2/converters/ConverterTestSuite.java
@@ -53,7 +53,6 @@ public class ConverterTestSuite {
         testSuite.addTestSuite(SqlDateConverterTestCase.class);
         testSuite.addTestSuite(SqlTimeConverterTestCase.class);
         testSuite.addTestSuite(SqlTimestampConverterTestCase.class);
-        testSuite.addTestSuite(URLConverterTestCase.class);
         return testSuite;
     }
 }
--- a/src/test/java/org/apache/commons/beanutils2/converters/URLConverterTestCase.java
+++ b/src/test/java/org/apache/commons/beanutils2/converters/URLConverterTestCase.java
@@ -17,57 +17,28 @@
 
 package org.apache.commons.beanutils2.converters;
 
+import org.apache.commons.beanutils2.ConversionException;
+import org.junit.Before;
+import org.junit.Test;
+
 import java.net.URL;
 
-import org.apache.commons.beanutils2.ConversionException;
-import org.apache.commons.beanutils2.Converter;
-
-import junit.framework.TestCase;
-import junit.framework.TestSuite;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 /**
  * Test Case for the URLConverter class.
- *
  */
+public class URLConverterTestCase {
 
-public class URLConverterTestCase extends TestCase {
+    private URLConverter converter;
 
-    public static TestSuite suite() {
-        return new TestSuite(URLConverterTestCase.class);
+    @Before
+    public void before() {
+        converter = new URLConverter();
     }
 
-
-
-    private Converter converter = null;
-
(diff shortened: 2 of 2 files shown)


melloware (approved):

jvz (approved):
Nice catch! The legacy behavior of `URL.equals()` is always a fairly hilarious bug to come across after you forget it exists for a few months or years. :)

melloware:

garydgregory:
> Can you bump your Jacoco in pom.xml 0.8.6 so it fixes the build and retriggers please?

Just rebase on master.

garydgregory:
If you have no internet, you won't be able to build anyway since Maven will download plugins to get the build off the ground.

melloware:
@garydgregory what commands do you use to rebased a banch to master.  Whenever I do it it makes my PR look like it has all those changes and muddies the main PR so I must be doing something wrong?

garydgregory:
> @garydgregory what commands do you use to rebase a branch to master. Whenever I do it it makes my PR look like it has all those changes and muddies the main PR so I must be doing something wrong?

git can be tricky to use, google will be your friend here.

melloware:
Ha yeah tried following: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request  I must have done something wrong.

SethFalco:
> If you have no internet, you won't be able to build anyway since Maven will download plugins to get the build off the ground.
> 

I appreciate this, however it's common to try minimise any reason a unit test fails other than from what the unit test is testing.

The URLConverter tests are only intended to test the conversion to the URL type. An internet connection, or if URLs resolve to the same host shouldn't be in scope for the test. Currently, it can fail when what we want to test is correct.

At the point one has already cloned the repo, and already resolved all dependencies and plugins, I firmly believe they shouldn't need internet to run unit tests locally.

SethFalco:
I've rebased the PR with `master` now, it should be up to date including the JaCoCo 0.8.6 commit.

SethFalco:
Looking back at this, I decided to refactor the class as well to avoid duplication.
I also added 2 more tests since I was checking results between `URL`/`URI`/`String`, I figured we could keep them in.

Please let me know if you believe the suggested changes are unfavorable.

src/test/java/org/apache/commons/beanutils2/converters/ConverterTestSuite.java
```diff
@@ -53,7 +53,6 @@ public static Test suite() {
         testSuite.addTestSuite(SqlDateConverterTestCase.class);
         testSuite.addTestSuite(SqlTimeConverterTestCase.class);
         testSuite.addTestSuite(SqlTimestampConverterTestCase.class);
-        testSuite.addTestSuite(URLConverterTestCase.class);
```
garydgregory, line 56: FYI: This whole class is no longer needed since we've been on Maven for a while... I'll remove it later.

PR #54, opened 2020-12-17
Minor Improvement

arturobernalg:
* Add Final
* change While loop

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/BeanMap.java
+++ b/src/main/java/org/apache/commons/beanutils2/BeanMap.java
@@ -45,9 +45,9 @@ public class BeanMap extends AbstractMap<String, Object> implements Cloneable {
 
     private transient Object bean;
 
-    private transient HashMap<String, Method> readMethods = new HashMap<>();
-    private transient HashMap<String, Method> writeMethods = new HashMap<>();
-    private transient HashMap<String, Class<? extends Object>> types = new HashMap<>();
+    private final transient HashMap<String, Method> readMethods = new HashMap<>();
+    private final transient HashMap<String, Method> writeMethods = new HashMap<>();
+    private final transient HashMap<String, Class<? extends Object>> types = new HashMap<>();
 
     /**
      * An empty array. Used to invoke accessors via reflection.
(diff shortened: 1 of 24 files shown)


melloware:
I am a big fan of this change and using the `final` keyword.  Joshua Bloch would approve!

melloware (approved):

garydgregory:
I think we need to get the build passing on Java 15 first to make it simpler to assess build results for PRs and master itself.

melloware:
@gary any hint at what is wrong?  It seems its JDK15 and higher that are the issue.

garydgregory:
> @gary any hint at what is wrong? It seems its JDK15 and higher that are the issue.

I only noticed the build failure, not its cause.

jochenw:
Are you reading? Rather than reducing the number of affected files, you just increased it to 24.

melloware:
> Are you reading? Rather than reducing the number of affected files, you just increased it to 24.

@jochenw What do you mean?  Reading what?  I don't understand what you are saying.

jochenw:
Sorry, mismatched this PR with another one on commons-fileupload. However, my opinion remains: 24 affected files is way too much for "minor improvements".

melloware:
Oh OK.  Yeah maybe the title is poor for this ticket but the improvements I think are good ones.

melloware:
@garydgregory I think the build is failing with "java.util.ServiceConfigurationError: Locale provider adapter "CLDR"cannot be instantiated."

Which according to Stack Overflow that Code Coverage is not supported in JDK13+ so that is what is failing.  I am surprised this is not happening on other Commons projects as well?

https://stackoverflow.com/questions/64696565/eclipse-testing-with-coverage-throws-java-lang-instrument-illegalclassformatexc

melloware:

arturobernalg:
Rebase from master.  
the change is basically add final to some var. It's complicated create a PR with only some change.

melloware:
Nice work build is passing on both your PR's now.

arturobernalg:
Thank you very mucho @melloware

arturobernalg:
Any idea of how can I split the PR?

melloware:
Create 2 new PR's and drop this one.

arturobernalg:
> Create 2 new PR's and drop this one.
Yep, the thing is, how many file? I think we need to have patience and give an opportunity to this kind of PR

melloware:
personally i would leave this PR.  its not that big.

PR #55, opened 2020-12-29
Jacoco 0.8.6 for JDK15 support

melloware:
Attempting to fix build as Jacoco 0.8.6 supports JDK15


melloware:
@garydgregory looks like this fixes the build!

PR #70, opened 2021-01-10
[BEANUTILS-525] - Enhance BeanUtils.copyProperties with ignore fields parameter

arturobernalg:
Copy the property values of the given source bean into the given target bean, ignoring the given "ignoreProperties".


arturobernalg:
HI @melloware 
What do u think about this new change?

PR #72, opened 2021-01-21
Refactor logger usage

korzha:
1. Skip message concatenation if logging level is not enabled
2. Remove logging level prefix from messages because it's printed by
logger formatter itself
3. Remove exception from message, pass it as a last parameter of slf4j
logger call


melloware (approved):

korzha:
@melloware Thanks for approval! 
But 17-ea build fails for some odd reason "Locale provider adapter "CLDR"cannot be instantiated."
Is there something missing in master?

garydgregory:
What about the 32 other calls to trace? What should we do (if anything) about those? If we are going to guard against wasteful string construction, let's do it consistently instead of one-off PRs.

korzha:
@garydgregory I've fixed all other logger misusages. I didn't touch single string parameter calls since they do level checks inside by themselves.

PR #80, opened 2021-03-03
Rename beanutils2 package to beanutils to be a drop-in replacement for beanutils 1.9.4

chrismaeda:
We had to eliminate commons-collections 3.x throughout our stack for a large corporation due to security vulnerabilities.

We used gradle dependency tricks to force this to be used instead of beanutils 1.9.4.

Could I help you release this as beanutils 1.9.5 ???


garydgregory:
-1 due to binary compatibility.

melloware:
-1 also would prefer a BeanUtils2 2.0.0 official release in Maven Central instead.

XenoAmess:
I suggest you fork and release on your host...

Just changing group ID will be fine...

After all this is apache2 license, right...

And ossrh is free..

melloware:
Yep I have done this before with other libraries and put in Maven Central in my `com.melloware` artifact but was really hoping not to have to do that with an Apache Commons Library.  But you are right I think I have no choice...

XenoAmess:
> Yep I have done this before with other libraries and put in Maven Central in my `com.melloware` artifact but was really hoping not to have to do that with an Apache Commons Library. But you are right I think I have no choice...

Yep...please only do this when necessary...

chrismaeda:
> > Yep I have done this before with other libraries and put in Maven Central in my `com.melloware` artifact but was really hoping not to have to do that with an Apache Commons Library. But you are right I think I have no choice...
> 
> Yep...please only do this when necessary...

So beanutils 1.9.4 is 2 years old and has a small dependency on commons-collections 3, which is red-flagged for security vulnerabilities.  A lot of things have dependencies on beanutils; e.g. Grails 4.0.x depends on commons-validator which depends on beanutils.

I'm offering to help do an update of these commons components to fix security issues.  But it sounds like the official position I'm getting here is that we should maintain our own forks and wait for version 2?

garydgregory:
> > > Yep I have done this before with other libraries and put in Maven Central in my `com.melloware` artifact but was really hoping not to have to do that with an Apache Commons Library. But you are right I think I have no choice...
> > 
> > 
> > Yep...please only do this when necessary...
> 
> So beanutils 1.9.4 is 2 years old and has a small dependency on commons-collections 3, which is red-flagged for security vulnerabilities. A lot of things have dependencies on beanutils; e.g. Grails 4.0.x depends on commons-validator which depends on beanutils.
> 
> I'm offering to help do an update of these commons components to fix security issues. But it sounds like the official position I'm getting here is that we should maintain our own forks and wait for version 2?

I would not call it an official position but more of a pragmatic view that we are all volunteering our time here and we all have different priorities. So the release will come when the component is ready and we will want to make sure that it is fully baked. One big issue the 1.x series had is that it surfaces in its API Common Collections types, such that one cannot switch to Collections 4 without breaking binary compatibility, hence one of the drivers for a new major version in a new package with new Maven coordinates.
> 
> Perhaps I should use the 1.9.4 source as a starting point instead??

That's what I would do.

chrismaeda:
> > > > Yep I have done this before with other libraries and put in Maven Central in my `com.melloware` artifact but was really hoping not to have to do that with an Apache Commons Library. But you are right I think I have no choice...
> > > 
> > > 
> > > Yep...please only do this when necessary...
> > 
> > 
> > So beanutils 1.9.4 is 2 years old and has a small dependency on commons-collections 3, which is red-flagged for security vulnerabilities. A lot of things have dependencies on beanutils; e.g. Grails 4.0.x depends on commons-validator which depends on beanutils.
> > I'm offering to help do an update of these commons components to fix security issues. But it sounds like the official position I'm getting here is that we should maintain our own forks and wait for version 2?
> 
> I would not call it an official position but more of a pragmatic view that we are all volunteering our time here and we all have different priorities. So the release will come when the component is ready and we will want to make sure that it is fully baked. One big issue the 1.x series had is that it surfaces in its API Common Collections types, such that one cannot switch to Collections 4 without breaking binary compatibility, hence one of the drivers for a new major version in a new package with new Maven coordinates.
> 
> BeanUtils 2 should not surface Commons Collections types in its APIs, which was one of our goals. BUT, we do not have a way to enforce that in our Maven build ATM, so I would like help on achieving that goal, such that a Maven build would fail if a Commons Collection type shows up in any public or protected type, method or field.
> 
> > Perhaps I should use the 1.9.4 source as a starting point instead??
> 
> That's what I would do.

Understood.  I see more clearly that this PR does not make sense for the master branch.  I'll attempt to duplicate my changes in the 1.X branch and submit a new PR against that.

garydgregory:
Closing, -1 above, see discussion.

melloware:
```xml
<dependency>
  <groupId>com.melloware</groupId>
  <artifactId>commons-beanutils2</artifactId>
  <version>2.0.0</version>
</dependency>
```
Is in Maven Central if anyone needs it.

PR #82, opened 2021-03-12
[BEANUTIL-547] MethodUtils java version check fix Java 11

spyrkob:
Issue: https://issues.apache.org/jira/browse/BEANUTILS-547


arturobernalg:
HI @spyrkob 
Could you provide a test to reproduce the problem?
TY

spyrkob:
Hi @arturobernalg updated the PR with a simple test case. The problem is visible when running with Java 11+

bstansberry:
Any news on this one? SE 8 is staring to fade into the background and any LTS SE that doesn't support security manager is two or three years away, and then likely will take many more years to displace SE 11 or 17. So utility libraries that don't work properly on a security manager on more recent SEs are problematic.

spyrkob:
Update the PR to resolve the conflict. Is there anything I can do to help get this fix merged?

PR #90, opened 2021-07-09
Update .gitignore to ignore Visual Studio Code files

SethFalco:
Similar to https://github.com/apache/commons-csv/pull/160

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,23 @@
-/target/
+### https://github.com/github/gitignore/blob/master/Maven.gitignore
+
+target/
+pom.xml.tag
+pom.xml.releaseBackup
+pom.xml.versionsBackup
+pom.xml.next
+release.properties
+dependency-reduced-pom.xml
+buildNumber.properties
+.mvn/timing.properties
+# https://github.com/takari/maven-wrapper#usage-without-binary-jar
+.mvn/wrapper/maven-wrapper.jar
+
+
 .classpath
 .project
 .settings/
+
 /.idea/
-/commons-beanutils2.iml
+*.iml
+
+/.vscode/


SethFalco:
Something to consider, it could be worth removing [CODE_OF_CONDUCT.md](https://github.com/apache/commons-beanutils/blob/master/CODE_OF_CONDUCT.md) from the root of the repository.
Two months ago, it was added to [apache/.github](https://github.com/apache/.github), so GitHub can infer it from there. ([More Info](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file
))

Also means when/if the file ever changes, it'll be done for you. No need to copy/paste it over everywhere.

This applies to [apache/commons-csv](https://github.com/apache/commons-csv) as well.

garydgregory:
Hi,

Does the file get included when you build our src zip for a release or
download a repo zip from GitHub?

Our md file just points to the Apache COC page, so there is no effective
duplication.

Gary

SethFalco:
> Does the file get included when you build our src zip for a release or
> download a repo zip from GitHub?

No, it just gets displayed in the GitHub UI as if it was in this repository as well.
(ie when a user tries to make a pull request)

Fair enough! Thanks for the response.

SethFalco:
I'll close this. I've since learned it's possible to lines to have a global `.gitignore` file, so there's no need to do it on a per-project basis.

If there are any reservations for a change like this as it's editor-specific, we can drop it! 👍🏽

TL;DR:

```sh
echo '/.vscode/
*.iml
' > ~/.gitignore
git config --global core.excludesFile ~/.gitignore

PR #92, opened 2021-07-19
Fix warnings and To-Dos

SethFalco:
While going through the project, I was just doing some minor refactors to tidy things up. Most of these were warnings from Sonarlint.

* `BeanComparator.java` - Switch final if/else to just return the result of the method.
* `BeanMap.java` - No need to declare a local variable just to return it immediately.
* `WrapDynaClass.java` - Iterate over the values, since the Entry key wasn't used.
* `SqlTimeConverter.java` / `SqlTimestampConverter.java` - Remove the TODOs and just inherit the documentation.
* `BeanMapTestCase.java` / `DynaBeanMapDecoratorTestCase.java` - Unnecessary @ SuppressWarnings
* `BeanificationTestCase.java` - Use more appropriate assertion method.
* `Jira357TestCase.java` - The test case passes when uncommented already.

Feel free to dispute any of the changes.


src/main/java/org/apache/commons/beanutils2/converters/SqlTimeConverter.java
```diff
@@ -68,9 +68,7 @@ public SqlTimeConverter(final Object defaultValue) {
     }
 
     /**
-     * Gets a {@code DateFormat} for the Locale.
-     * @param locale TODO
-     * @param timeZone TODO
+     * {@inheritDoc}
```
garydgregory, line 71: Hi. You removed some of the tags but not the return tag, why some and not that one?

src/main/java/org/apache/commons/beanutils2/converters/SqlTimestampConverter.java
```diff
@@ -68,9 +68,7 @@ public SqlTimestampConverter(final Object defaultValue) {
     }
 
     /**
-     * Gets a {@code DateFormat} for the Locale.
-     * @param locale TODO
-     * @param timeZone TODO
+     * {@inheritDoc}
```
garydgregory, line 71: Same question as above.

garydgregory (changes requested):
@SethFalco 
TY for your PR; please see my comments.

src/main/java/org/apache/commons/beanutils2/converters/SqlTimeConverter.java
SethFalco: Valid 🤔 

I remember at the time thinking I'll leave the `@return` tag because it was different, so I assumed it should be overridden.
Now looking at it... I realize it's "different" but not actually different.

I'll remove it. ^-^'

SethFalco:
I removed the `@return` and `{@inheritDoc}`.

I was checking into it, but seems `inheritDoc` is only useful to inherit and extend the documentation body.
Didn't realize the tags (`@param`, `@return`, etc) would still get inherited by default even though we're defining new tags.

SethFalco:
Just rebased with master, hoping this should be cool now.
Seems I forgot to re-request review after resolving the comments, sorry about that. ^-^'

melloware (approved):

SethFalco:
I've just rebased the changes with `master` and resolved the merge conflicts. All tests pass for me locally too.

src/main/java/org/apache/commons/beanutils2/sql/converters/SqlTimeConverter.java
```diff
@@ -62,12 +62,6 @@ protected Class<Time> getDefaultType() {
     }
 
     /**
-     * Gets a {@code DateFormat} for the Locale.
```
garydgregory, line 65: This seems to be going in the wrong direction. You want to fill in the TODO, not blow away the whole Javadoc comment.

src/main/java/org/apache/commons/beanutils2/sql/converters/SqlTimestampConverter.java
```diff
@@ -62,12 +62,6 @@ protected Class<Timestamp> getDefaultType() {
     }
 
     /**
-     * Gets a {@code DateFormat} for the Locale.
```
garydgregory, line 65: As above.

src/test/java/org/apache/commons/beanutils2/BeanificationTestCase.java
```diff
@@ -327,9 +329,21 @@ public String toString() {
         thread.join();
 
         assertEquals("Signal not set by test thread", 2, signal.getSignal());
-        assertTrue("Different BeanUtilsBean instances per context classloader", BeanUtilsBean.getInstance() != signal.getBean());
-        assertTrue("Different ConvertUtilsBean instances per context classloader", ConvertUtilsBean.getInstance() != signal.getConvertUtils());
-        assertTrue("Different PropertyUtilsBean instances per context classloader", PropertyUtilsBean.getInstance() != signal.getPropertyUtils());
+        assertNotEquals(
```
garydgregory, line 332: What's the change here? If it's only formatting, please revert, it just makes PRs noisier.

garydgregory (changes requested):
Hello @SethFalco 
Thank you for the update.
Please see my comments.

src/main/java/org/apache/commons/beanutils2/sql/converters/SqlTimestampConverter.java
SethFalco: It's already inheriting the JavaDoc from the function it's overriding. I removed it because it didn't look like it needed populating.

image

src/test/java/org/apache/commons/beanutils2/BeanificationTestCase.java
SethFalco: Before: Used `assertTrue` with a condition to it.  
After: Uses `assertNotEquals` with the 2 variables passed to it.

Typically, with JUnit `assertEquals` or `assertNotEquals` is used rather than `==` or `!=`.

Happy to turn it back to a single line though, but the primary change was the assertion method that was used.

PR #56, opened 2020-12-29
BEANUTILS-539/BEANUTILS-509: Switch WeakFastHashMap for ConcurrentWeakHashMap

melloware:
Gave attribution to original authors and the original code has this disclaimer:

_Written by Doug Lea with assistance from members of JCP JSR-166
Expert Group and released to the public domain, as explained at_
http://creativecommons.org/licenses/publicdomain

All unit tests passing.

PR #95, opened 2021-09-01
BEANUTILS-539/BEANUTILS-509: 1.X line, switch WeakFastHashMap for ConcurrentWeakHashMap 

car-roll:
Porting the changes made in #56 to the 1.X line


olamy:
@garydgregory this issue makes Jenkins totally irresponsive... see https://issues.jenkins.io/browse/JENKINS-60997
such race condition which depends on the garbage collector (because of weakreference) is definitely impossible to reproduce  100% with a unit test.
what could be your proposal alternative to fix that?

garydgregory:
> @garydgregory this issue makes Jenkins totally irresponsive... see https://issues.jenkins.io/browse/JENKINS-60997
> such race condition which depends on the garbage collector (because of weakreference) is definitely impossible to reproduce 100% with a unit test.
> what could be your proposal alternative to fix that?

I expect the new classes to be tested thoroughly. I can understand that the whole use case might be hard to reproduce but a whole new class needs a whole new test to match it. Public and protected APIs cannot be changed.

melloware:
Now that 2.0.0-M1 is out you can officially close this.

PR #96, opened 2021-09-06
Refactor IntrospectionContextTestImpl in SuppressPropertiesBeanIntros…

wx930910:
Fix [BEANUTILS-551](https://issues.apache.org/jira/browse/BEANUTILS-551)

#### Replace test class [IntrospectionContextTestImpl](https://github.com/apache/commons-beanutils/blob/7196d86767e37a382329bb3b087d0f58c7301caf/src/test/java/org/apache/commons/beanutils2/SuppressPropertiesBeanIntrospectorTestCase.java#L101) by mocking object and improve test design
<hr>

##### Motivation

- Decouple test class `IntrospectionContextTestImpl` from production interface `IntrospectionContext`.
- Extract attribute into test cases to make test condition more explict.
- Make test logic more clear by using method stub instead of method overriding.

<hr>

##### Key changed/added classes in this PR
 * Created mocking object to replace test subclass `IntrospectionContextTestImpl`, decoupled test from production code.
 * Extract `removedProperties` as a local variable to improve test logic and make test condition more explict.
 * Make test logic more clear by using method stub instead of method overriding.
 * Add Mockito dependency

<hr>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -350,6 +350,13 @@
       <version>4.13.2</version>
       <scope>test</scope>
     </dependency>
+	<!-- https://mvnrepository.com/artifact/org.mockito/mockito-core -->
+	<dependency>
+	    <groupId>org.mockito</groupId>
+	    <artifactId>mockito-core</artifactId>
+	    <version>3.9.0</version>
+	    <scope>test</scope>
+	</dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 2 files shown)


melloware (approved):

garydgregory:
I'm going to say no here for the same reason as for the similar PRs proposed in other Commons components: this makes the code harder to follow and maintain and does not offer benefits.

PR #106, opened 2022-02-14
BEANUTILS-550: correction of returning default value by StringConverter

JKLedzion:
PR for Jira issue: BEANUTILS-550.

Assignment of default value in StringConverter didn’t work as method getDefault(final Class<?> type), only for String.class, returned null instead of default value. I’ve changed this method to return default value also for String.class.
 
Could you please review and let me know if this modification is OK? 
I would appreciate any feedback.
Justyna


garydgregory:
@JKLedzion 
Thank you for the PR. Don't use Javadoc comments for comments that are not Javadoc. Inline comments would be more appropriate.

JKLedzion:
> @JKLedzion Thank you for the PR. Don't use Javadoc comments for comments that are not Javadoc. Inline comments would be more appropriate.

Corrected.

PR #196, opened 2024-01-01
BEANUTILS-560: build multi-release jar

SingingBush:


garydgregory:
This should not be needed since we use the moditect maven plugin through the parent POM.

garydgregory:
Closing: We use the Moditect plugin to generate JPMS artifacts.

PR #201, opened 2024-01-16
fix for JDK 21 CI failure due to locale changes

SingingBush:


garydgregory:
@SingingBush 
Please always use the description in a PR, add a URL to the JIRA ticket when there is one. TY! 😄

PR #232, opened 2024-04-08
Change visibility of PropertyUtilsBean.getReadMethod(Class, PropertyDescriptor)

doublep:
Having this method package-private (as now) hinders subclassing. This cannot be considered a security issue, because nothing precludes subclasses from overriding every public method where the method in question is used in anyway.

This change is also consistent with visibility of other similar methods. For comparison:

* `getReadMethod(Class)` — public;
* `getReadMethod(Class, PropertyDescription)` — currently package-private, changing to public;
* `getWriteMethod(Class)` — public;
* `getWriteMethod(Class, PropertyDescription)` — public (added only in 1.9.1).

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java
+++ b/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java
@@ -1119,6 +1119,7 @@ public class PropertyUtilsBean {
      * @param clazz The class of the read method will be invoked on
      * @param descriptor Property descriptor to return a getter for
      * @return The read method
+     * @since 2.0.0
      */
     public Method getReadMethod(final Class<?> clazz, final PropertyDescriptor descriptor) {
         return MethodUtils.getAccessibleMethod(clazz, descriptor.getReadMethod());
--- a/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java
+++ b/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java
@@ -1120,7 +1120,7 @@ public class PropertyUtilsBean {
      * @param descriptor Property descriptor to return a getter for
      * @return The read method
      */
-    Method getReadMethod(final Class<?> clazz, final PropertyDescriptor descriptor) {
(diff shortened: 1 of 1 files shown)


garydgregory:
If you intend to use it in a subclass, why is it public and not protected?

doublep:
For consistency with getWriteMethod(Class, PropertyDesciptor). But I don't
mind if it is made protected, just not package-private as now.

Paul

doublep:
I just checked the source code history and this is the commit where `getWriteMethod(Class, PropertyDescriptor)` was made public:

    8f77bd63 [BEANUTILS-456] Provide another method for querying a write method.
    Author:     Oliver Heger <oheger@apache.org>
    AuthorDate: 2014-01-03 19:38:03 +0000

So, changing it back (or to `protected`) after ten years would be a bit backward-incompatible. So, instead I suggest to follow suit and do _the same_ for `getReadMethod(Class, PropertyDescriptor)`.

garydgregory:
@doublep 
We can break BC in a major release like 2.0.0. The question really is: What is the best design? WDYT?

doublep:
I'd say keep all four `getReadMethod()` and `getWriteMethod()` overloads public. As these method names are very common, searching in publicly available code doesn't yield useful results (too many unrelated hits), but suppose someone somewhere has a valid usecase to access one of these methods. (I can't think of one myself, but I don't rule out the possibility that there might be.)

If you now lower the visibility, you will break someone's code for no particularly good reason, forcing them to look for workarounds. E.g. subclassing, making the methods public and then explicitly downcasting to the created class when using from "outside". Basically, a lot of hassle for virtually nothing. It's not like `PropertyUtilsBean` is a high-level class where having a nice compact interface is a lot of value.

So, I'd leave them all `public`, but add a "disclaimer" in the documentation that these methods are lower-level than the rest (e.g. `get*Property()`) and should generally not be needed outside of some special usecases.

garydgregory:
@doublep Do you want to update the Javadoc as part of this PR or a separate PR?

doublep:
@garydgregory I just pushed some Javadoc changes. Feel free to rewrite or adjust as you prefer, easier this way than negotiating in advance.

PR #267, opened 2024-07-30
BEANUTILS-566: Use ContextClassLoader for Instantiating Loggers

volosied:
https://issues.apache.org/jira/browse/BEANUTILS-566

Swaps our the  thread context classloader to avoid memory leaks in some scenarios.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/bugs/Jira566TestCase.java
+++ b/src/test/java/org/apache/commons/beanutils2/bugs/Jira566TestCase.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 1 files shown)


garydgregory:
20 files changed and zero tests is not good.

Try and see if Mockito helps.

volosied:
Hi @garydgregory , 

I've found a way to test that the context classloader is used. I also tested the negative case to verify it's not called.  There may be an easier way to test, but I simply created my own classloader and load the tests from the target/test-classes directory. 

Classloading isn't my expertise, and my teammate who is had recently left.  This is a good learning experience. 

I'm not sure how to run the jobs -- they failed due to checkstyle violations which I've since fixed. 

I would appreciate any feedback. Thank you.

volosied:
I pushed up another test which also let the builds run.

Everything passed except for Java 23, but those failures are unrelated to my changes from the looks of. 
 
```
 Error:  Failures: 
Error:    SqlTimeConverterTestCase.testLocale:118->AbstractDateConverterTest.validConversion:572 Converting 'java.lang.String' value '3:06 pm' threw org.apache.commons.beanutils2.ConversionException: Error converting 'String' to 'java.sql.Time' using pattern 'h:mm a', localized pattern 'h:mm a', errorIndex 4, calendar type GregorianCalendar, this org.apache.commons.beanutils2.sql.converters.SqlTimeConverter[UseDefault=false, UseLocaleFormat=true]
Error:    SqlTimestampConverterTestCase.testLocale:138->AbstractDateConverterTest.validConversion:572 Converting 'java.lang.String' value '3/21/06, 3:06 PM' threw org.apache.commons.beanutils2.ConversionException: Error converting 'String' to 'java.sql.Timestamp' using pattern 'M/d/yy, h:mm a', localized pattern 'M/d/yy, h:mm a', errorIndex 13, calendar type GregorianCalendar, this org.apache.commons.beanutils2.sql.converters.SqlTimestampConverter[UseDefault=false, UseLocaleFormat=true]
[INFO] 
```

ppkarwasz (changes requested):
@volosied,

IMHO the TCCL problem should be solved at the source in Commons Logging.
See [LOGGING-196](https://issues.apache.org/jira/browse/LOGGING-196) for details.

PR #276, opened 2024-08-22
BeanUtils.setProperty may shouln't record trace log

F0otman:
From [BEANUTILS-568](https://issues.apache.org/jira/browse/BEANUTILS-568),
Sorry for lacked unit test, cause beanutils only use commons-logging, it cannot show the key point of log without settings or materialized log component.

The problem I want to warn is if I override bean method toString like
`return "TestBean{" + "propertyPassword='" + "******" + "\" + '}'`
When try to use BeanUtils.setProperty & BeanUtils.copyProperty & LocaleBeanUtilsBean.setProperty, **it would record trace log the value of propertyPassword**. So if someone tamper log level or set the wrong level, the shaded property may record in log.

The safety way is when toString method is overrided, do not show the value in log.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -72,7 +72,7 @@
         Refactor logger usage #72.
       </action>
       <action dev="ggregory" type="fix" due-to="SethFalco, Steve Bosman, Gary Gregory">
-        Migrate to JUnit 5 #93, #283, #284, #285.
+        Migrate to JUnit 5 #93, #283, #284, #285, #287.
       </action>
       <action dev="ggregory" type="fix" due-to="Gary Gregory">
         Fix SpotBugs [ERROR] Medium: org.apache.commons.beanutils2.BasicDynaClass.constructorTypes should be both final and package protected [org.apache.commons.beanutils2.BasicDynaClass] At BasicDynaClass.java:[line 95] MS_FINAL_PKGPROTECT.


garydgregory:
@F0otman 

Would you like to use the new method from your code? If not, please make the method private.

F0otman:
The method not only use for BeanUtilsBean.copyProperty and BeanUtilsBean.setProperty, but also for LocaleBeanUtilsBean.setProperty, they are same case and duplicate code. Protected id better

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
```diff
@@ -380,29 +380,7 @@ public void copyProperty(final Object bean, String name, Object value)
...
-                    }
-                    sb.append(values[i]);
-                }
-                sb.append(']');
-            } else {
-                sb.append(value.toString());
-            }
-            sb.append(')');
-            LOG.trace(sb.toString());
+            LOG.trace(traceLogRecord(bean, name, value, sb).toString());;
```
garydgregory, line 383: Too many semicolons;;

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
```diff
@@ -911,29 +889,7 @@ public void setProperty(final Object bean, String name, final Object value)
...
```
garydgregory, line 892: Too many semicolons;;

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
```diff
@@ -1067,4 +1023,57 @@ public void setProperty(final Object bean, String name, final Object value)
...
+     * {@code bean.toString}, property name and property value.</p>
+     *
+     * <p><strong>When the bean's toString method is override, hide the detail
+     * of value。</strong></p>
+     *
+     * @param bean Bean on which setting is to be performed
+     * @param name Property name (can be nested/indexed/mapped/combo)
+     * @param value Value to be set
+     */
+    protected static StringBuilder traceLogRecord(Object bean, String name, Object value, StringBuilder sb) {
```
garydgregory, line 1039: Missing `@param` tag.
Missing `@return` tag.

src/main/java/org/apache/commons/beanutils2/locale/LocaleBeanUtilsBean.java
```diff
@@ -796,32 +796,7 @@ public void setProperty(
...
-                    sb.append(values[i]);
-                }
-                sb.append(']');
-            }
-            else {
-                sb.append(value.toString());
-            }
-            sb.append(')');
-            LOG.trace(sb.toString());
+            LOG.trace(traceLogRecord(bean, name, value, sb).toString());;
```
garydgregory, line 799: Too many semicolons;;

garydgregory (changes requested):
Hello @F0otman 
Please see my comments.

F0otman:
Thanks for the comments.
I heared beanutils 2 is coming, any feature I could supplement?

garydgregory:
> So if someone tamper log level or set the wrong level, the shaded property may record in log.

If some has this kind of access to your application, you're already in very deep trouble. Note that some folks recommend to never store credentials in `String`s for this reason, and use `char[]` instead.

Let's see if anyone else has feedback.

garydgregory:
> Thanks for the comments. I heared beanutils 2 is coming, any feature I could supplement?

One remaining development task for 2.x is to pull up the "2" types
into their parent:

- org.apache.commons.beanutils2.BeanUtilsBean2 extends BeanUtilsBean
- org.apache.commons.beanutils2.ConvertUtilsBean2 extends ConvertUtilsBean

The last time I tried this, tests failed and I was semi-stuck as to
how to fix those.

See also:
- https://issues.apache.org/jira/browse/BEANUTILS-565
- https://issues.apache.org/jira/browse/BEANUTILS-569

Discussions and PRs on GitHub are welcome.

sigee:
I think a couple of test cases would be nice, at least for the new logic.

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
```diff
@@ -1067,4 +1023,59 @@ public void setProperty(final Object bean, String name, final Object value)
...
+     * <p><strong>When the bean's toString method is override, hide the detail
+     * of value.</strong></p>
+     *
+     * @param bean Bean on which setting is to be performed
+     * @param name Property name (can be nested/indexed/mapped/combo)
+     * @param value Value to be set
+     * @param sb the start of stringBuilder
+     * @return the value should log as trace
+     */
+    protected static StringBuilder traceLogRecord(Object bean, String name, Object value, StringBuilder sb) {
```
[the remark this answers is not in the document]
F0otman: Your suggestion may right, the futher code I wrote tried to keep as same as original possibly.
Your's are more sence

F0otman:
Test cases must by fine, but the changed code is use for logging. Beanutils only use commons-logging as logging adapter system, I cannot directly log without import any other implementations logging software (At least I have no idea).

Anyone have idea?

sigee:
@F0otman 
Typically, logging is not a feature you have to test, but with this change, you put a little "business logic" into it.

About the test writing. You are right, it is a bit complicated because the code may not be in a testable format.
"Never hide a TUF within a TUC."
TUF - Test Unfriendly Feature
- a piece of functionality embodied in code, which makes unit testing difficult
- database access
- filesystem access
- network access
- static variable usage

TUC - Test Unfriendly Construct
- final methods
- final classes
- static methods
- private methods
- static initialization expressions

I'll think about it and will be back tomorrow.
Maybe a small refactor (create a constructor where you can inject a LOG) and you can assert the LOG.trace() was called with the expected parameter.

sigee:
I created a PoC commit [here](https://github.com/sigee/commons-beanutils/commit/fb1f2171e3afd66b01606a422d3ac7cc5309dc2c).
Not my best work, but I thought about something like this.

- I removed the final modifyer from LOG and created a constructor to inject my spy object.
- In the test class I created 3 test cases
  - Basic case
  - Null case
  - toString() is overriden
- I also created the LogSpy to check the parameters trace was called with.
- I also created a child test bean where I override toString()

For the spy Mockito would be great, because this hand-made spy feels wierd. But Mockito is not in the project and I'm not sure why.

Anyway. What do you guys think? Is is a way to go or I just overengineering it?

melloware:
+1 on this change in general though as we have been burned by stuff like this in the past someone logging a password field in a toString.

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
```diff
@@ -1067,4 +1039,60 @@ public void setProperty(final Object bean, String name, final Object value)
                 (e, "Cannot set " + propName);
         }
     }
+
+    /**
+     * <p>Build the stringBuilder by using set/copy bean property for log, only
```
garydgregory, line 1044: This comment is not quite right IMO: It mentions logging but the method does not log or check log levels. Since it is a protected method, it can be called from a subclass so you have no idea how it will be used. All of this to say, I think the Javadoc should just say what it does.

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
```diff
@@ -172,6 +172,22 @@ public BeanUtilsBean(final ConvertUtilsBean convertUtilsBean,
...
+     * to perform conversions from one object to another
+     * @param propertyUtilsBean use this {@code PropertyUtilsBean}
+     * to access properties
+     * @param log use this {@code Log}
+     * to appoint logging
+     */
+    public BeanUtilsBean(final ConvertUtilsBean convertUtilsBean, final PropertyUtilsBean propertyUtilsBean, final Log log) {
+        this.convertUtilsBean = convertUtilsBean;
+        this.propertyUtilsBean = propertyUtilsBean;
+        LOG = log;
```
garydgregory, line 188: This looks very bad IMO: Every time an instance is created, a global variable is edited, which sounds to me like a recipe for bugs and libraries blaming each other.

garydgregory (changes requested):
Hello @F0otman 

Please see my comment.

src/main/java/org/apache/commons/beanutils2/BeanUtilsBean.java
sigee: Not every time. The original constructors have not changed, they do not edit the global LOG. But you have to have some edit option if you want to inject dependency for mocking. It could be a constructor or a setter. For my PoC this was easyer, but as I said, not my best work. It could be much better if someone have time to really think about the solution. Unfortunately I can't work on it now, because I only have my phone for 2 weeks as technology.

F0otman:
Sorry for wrong git operate as closed, I synced fork master from here. I would pull a new pr these days :-(

PR #278, opened 2024-08-28
[BEANUTILS-565] merge BeanUtilsBean2, ConvertUtilsBean2 up

bmarwell:
This PR actually removed the delegation:

```
    @Override
    public <T> Object convert(final String[] value, final Class<T> clazz) {
        return convert((Object) value, clazz);
    }
```

This method from `ConvertUtilsBean2` was never under test and may never have worked.
However, the "old" method uses a lookup which is untyped. I am not sure whether the correct type `<T>` is returned in all instances.


melloware (approved):

bmarwell:
@melloware we need to fix those tests first... Input is welcome

melloware:
OK this is weird the error is 

```
 expected:<[28.12.]1992> but was:<[Mon Dec 28 00:00:00 UTC ]1992>
```

But its converting date to toString so I am not sure why it was ever expecting "28.12.1992" ???

bmarwell:
Good question. Does this test depend on some locale? That's a date commonly written like this here in Europe. Merging those two, I might have made an mistake. Also, the new methods delegate to different implementations now, which behave differently. 

Without knowing what the previous/original expectations of those methods were, it's really hard to tell. 

Will take a look at the docs and Jira issues and git history soon, maybe you can find something out, too?

melloware:
OK yes let me look it would almost seem like some Converter was at play that it would expect a date format like that which to me is unexpected.

garydgregory:
Please see the email thread I started on the dev mailing list:
https://lists.apache.org/thread/lg1xs63gjkm0v28nm4d2xsv5vgdq2436

garydgregory:
Hello @bmarwell , @melloware 

The pull-up refactoring is now in git master will tests adjusted for the "2" behaviors.
The "2" classes have been removed.
Please pull git master and verify your use cases.
I think this PR can be closed unless I missed something 😉 

TY!

melloware:
Yep I think it can be closed!

garydgregory:
Closing, no more feedback.

PR #283, opened 2024-08-31
Migrate to JUnit 5

stevebosman:
I've updated a lot of the remaining tests to JUnit 5

I think this will leave the tests extending the following to do:

* `AbstractDateConverterTest`
* `AbstractNumberConverterTest`
* `BaseLocaleConverterTest`

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/WrapDynaBeanTestCase.java
+++ b/src/test/java/org/apache/commons/beanutils2/WrapDynaBeanTestCase.java
@@ -17,10 +17,22 @@
 
 package org.apache.commons.beanutils2;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 
 /**
  * <p>
@@ -28,19 +40,7 @@ import java.io.ObjectOutputStream;
  * provide similar levels of functionality.
  * </p>
  */
-
 public class WrapDynaBeanTestCase extends BasicDynaBeanTestCase {
-
-    /**
-     * Constructs a new instance of this test case.
-     *
-     * @param name Name of the test case
-     */
-    public WrapDynaBeanTestCase(final String name) {
-
-        super(name);
-
-    }
 
     /**
      * Helper method for testing whether basic access to properties works as expected.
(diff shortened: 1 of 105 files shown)


garydgregory (changes requested):
Hello @stevebosman 

Thank you for your PR. 

It will take a lot longer to review a PR when it contains unnecessary changes like editing public/protected/private. Also please keep the style to use static imports only for assertion methods. This will all reduce the noise in the change set a lot.

IMO a PR for this task is not about creating the prettiest code but creating a change set that easier for a maintainer to review, and there are 105 files changed here 😉

stevebosman:
I've added the changes for static imports. I've also reverted the visibility changes. I think I've caught all of them, but will have another look later. My fault for trusting my IDE's refactoring too much :)

garydgregory:
@stevebosman 

Thank you for your update. Yeah, all IDEs I've tried mess up the conversion or leaves you with code that does not even compile (I'm looking at you IntelliJ...)

Another item that would be nice to have (but don't feel obligated):
```
assertTrue(e.toString().contains("Unknown property"), "Wrong exception was thrown: " + e);
```
->
```
assertTrue(e.toString().contains("Unknown property"), () -> "Wrong exception was thrown: " + e);
```
which will avoid creating garbage Strings in the majority of build runs.

stevebosman:
I have fixed that one and the few others in `BeanComparatorTestCase`. There are similar in some other test files, but my instinct is they should be addressed in another much smaller PR.

src/test/java/org/apache/commons/beanutils2/WrapDynaBeanTestCase.java
```diff
@@ -266,10 +276,10 @@ public void testMappedRemove() {
         }
 
         try {
-            assertTrue("Can not see unknown key", !bean.contains("mappedProperty", "Unknown Key"));
+            assertFalse(bean.contains("mappedProperty", "Unknown Key"), "Can not see unknown key");
             bean.remove("mappedProperty", "Unknown Key");
             fail("Should have thrown UnsupportedOperationException");
-            // assertTrue("Can not see unknown key",
+            // Assert.assertTrue("Can not see unknown key",
```
garydgregory, line 282: Odd comment change ;-)

src/test/java/org/apache/commons/beanutils2/bugs/Jira157TestCase.java
```diff
@@ -89,16 +84,16 @@ protected void setUp() throws Exception {
      *
      * @throws Exception
      */
-    @Override
+    @AfterEach
     protected void tearDown() throws Exception {
-        super.tearDown();
     }
```
garydgregory, line 89: So this method can be removed?

src/test/java/org/apache/commons/beanutils2/bugs/Jira273TestCase.java
```diff
@@ -16,54 +16,48 @@
...
-
     /**
      * Sets up.
      *
      * @throws Exception
      */
-    @Override
+    @BeforeEach
     protected void setUp() throws Exception {
-        super.setUp();
```
garydgregory, line 51: Remove method?

src/test/java/org/apache/commons/beanutils2/bugs/Jira273TestCase.java
```diff
@@ -16,54 +16,48 @@
...
 
     /**
      * Tear Down.
      *
      * @throws Exception
      */
-    @Override
+    @AfterEach
     protected void tearDown() throws Exception {
-        super.tearDown();
```
garydgregory, line 61: Remove method?

src/test/java/org/apache/commons/beanutils2/bugs/Jira298TestCase.java
```diff
@@ -16,56 +16,50 @@
...
-
     /**
      * Sets up.
      *
      * @throws Exception
      */
-    @Override
+    @BeforeEach
     protected void setUp() throws Exception {
-        super.setUp();
```
garydgregory, line 53: Remove method?

src/test/java/org/apache/commons/beanutils2/bugs/Jira298TestCase.java
```diff
@@ -16,56 +16,50 @@
...
 
     /**
      * Tear Down.
      *
      * @throws Exception
      */
-    @Override
+    @AfterEach
     protected void tearDown() throws Exception {
-        super.tearDown();
```
garydgregory, line 63: Remove method?

garydgregory (approved):
LGTM, I made a few notes, but I'll merge as is.

PR #314, opened 2024-11-14
create method copyNonNullProperties

XiongDWM:
When updating an entity, only a subset of properties may be provided (e.g., For a person entity[name:String, age:Integer, gender:boolean,] while only the name property is provided, or says we only want to update person's name, and other properties like age are null). In such cases, copying all properties including null values will overwrite the existing values in the target object, which is not desirable. This feature will prevent null values from overwriting existing values in the target object, ensuring that only the provided properties are updated.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/BeanUtils.java
+++ b/src/main/java/org/apache/commons/beanutils2/BeanUtils.java
@@ -111,6 +111,23 @@ public class BeanUtils {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


F0otman:
how about expand method doesn't copy empty properties as well

XiongDWM:
> how about expand method doesn't copy empty properties as well

you mean properties like empty string or empty collection? sure tho

PR #364, opened 2025-08-11
Simplify `BooleanConverter` and `ColorConverter`

basil:
See #359 for context. commit 53cd215292bd19ec6a467156eea031fe83c1defc from 2005 added a performance optimization that makes this code quite a bit more complicated than it needs to be. This strikes me as an example of premature optimization. The code can be made half as long (removing 33 lines) by simplifying this to use the standard `.equalsIgnoreCase` method. This also makes the code simpler to read and easier to reason about. An updated test (which passes both before and after the changes to `src/main`) confirms that this simplified code works for unusual locales.

Before you push a pull request, review this list:

- [x] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) if you use Artificial Intelligence (AI).
- [ ] I used AI to create any part of, or all of, this pull request.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best-practice.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/ColorConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/ColorConverter.java
@@ -90,42 +90,38 @@ public class ColorConverter extends AbstractConverter<Color> {
         if (Color.class.isAssignableFrom(type)) {
             final String stringValue = toString(value);
 
-            switch (toLowerCase(stringValue)) {
-            case "black":
+            if (stringValue.equalsIgnoreCase("black")) {
                 return type.cast(Color.BLACK);
-            case "blue":
+            } else if (stringValue.equalsIgnoreCase("blue")) {
                 return type.cast(Color.BLUE);
-            case "cyan":
+            } else if (stringValue.equalsIgnoreCase("cyan")) {
                 return type.cast(Color.CYAN);
-            case "darkgray":
-            case "darkgrey":
-            case "dark_gray":
-            case "dark_grey":
+            } else if (stringValue.equalsIgnoreCase("darkgray")
+                    || stringValue.equalsIgnoreCase("darkgrey")
+                    || stringValue.equalsIgnoreCase("dark_gray")
+                    || stringValue.equalsIgnoreCase("dark_grey")) {
                 return type.cast(Color.DARK_GRAY);
-            case "gray":
-            case "grey":
+            } else if (stringValue.equalsIgnoreCase("gray") || stringValue.equalsIgnoreCase("grey")) {
                 return type.cast(Color.GRAY);
-            case "green":
+            } else if (stringValue.equalsIgnoreCase("green")) {
                 return type.cast(Color.GREEN);
-            case "lightgray":
-            case "lightgrey":
-            case "light_gray":
-            case "light_grey":
+            } else if (stringValue.equalsIgnoreCase("lightgray")
+                    || stringValue.equalsIgnoreCase("lightgrey")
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/beanutils2/converters/ColorConverter.java
```diff
@@ -90,42 +90,38 @@ protected <T> T convertToType(final Class<T> type, final Object value) throws Th
         if (Color.class.isAssignableFrom(type)) {
             final String stringValue = toString(value);
 
-            switch (toLowerCase(stringValue)) {
-            case "black":
+            if (stringValue.equalsIgnoreCase("black")) {
                 return type.cast(Color.BLACK);
-            case "blue":
+            } else if (stringValue.equalsIgnoreCase("blue")) {
```
garydgregory, line 95: -1 This is terrible IMO for two reasons: 1) the one-step switch logic has devolved into a cascading if-else and 2) equals() is replaced by the ignore case version. This is worse all for catering to your needs. Previously the constructor performed a conversion once, now you've got equalsIgnoreCade all over the place.

ppkarwasz: I think we first need to clarify **what** we’re trying to achieve here, because Unicode case-insensitive comparison is not a trivial problem.

If our goal is to support *localized* color names, neither `toLowerCase(...)` nor `equalsIgnoreCase(...)` is sufficient for true case-insensitive matching.

For example, in German:

```java
final String left = "WEISS";
final String right = "weiß";
LOGGER.info("Comparing German words: left='{}', right='{}'", left, right);

LOGGER.info("Upper case (Locale.ROOT): left='{}', right='{}'",
        left.toUpperCase(Locale.ROOT),
        right.toUpperCase(Locale.ROOT));

LOGGER.info("Lower case (Locale.ROOT): left='{}', right='{}'",
        left.toLowerCase(Locale.ROOT),
        right.toLowerCase(Locale.ROOT));

LOGGER.info("String.equalsIgnoreCase: {}", left.equalsIgnoreCase(right));

final Collator collator = Collator.getInstance(Locale.ROOT);
collator.setStrength(Collator.PRIMARY);
LOGGER.info("Collator (PRIMARY strength): {}", collator.equals(left, right));
```

Output:

```
Comparing German words: left='WEISS', right='weiß'
Upper case (Locale.ROOT): left='WEISS', right='WEISS'
Lower case (Locale.ROOT): left='weiss', right='weiß'
String.equalsIgnoreCase: false
Collator (PRIMARY strength): true
```

Here, `"weiß"` and `"WEISS"` differ only in case, yet:

* `toLowerCase` fails because it preserves `ß`.
* `equalsIgnoreCase` fails because it compares per character without case folding.
* Only a full case folding comparison (`Collator` with `PRIMARY` strength) treats them as equal.

If we really want proper locale support, we need to use the right tool for the job, `Collator`, not just string case transformations.

basil: > I think we first need to clarify **what** we’re trying to achieve here, because Unicode case-insensitive comparison is not a trivial problem.

@ppkarwasz Thanks for asking! As stated in the PR description, this is a code simplification PR. The goal I am trying to achieve is a net simplification without any regressions in correctness and without (significant) regressions in performance.

> If our goal is to support localized color names

That is not the goal in this PR. The `java.awt.Color` has no knowledge of locales, and I see no compelling reason to support color translation as part of this library.

PR #398, opened 2026-06-05
Do not initialize arbitrary classes in EnumConverter

digi-scrypt:
1. convertToType resolves a fully qualified name straight from the conversion input with Class.forName(name), which both loads and initializes the named class before the isEnum() check runs.
2. so a String coming from an untrusted source (for example a request parameter bound through BeanUtils onto an enum property) can name any class on the classpath and fire its static initializer, even though the conversion then fails with "isn't an enumerated type".

Switched to the three-arg Class.forName(name, false, loader) so the class is resolved but not initialized until Enum.valueOf actually accepts it. Enum resolution is unchanged.

What happens with a hostile value: `converter.convert(Enum.class, "com.example.Gadget#X")` runs Gadget's static block today; after the change it does not. Have we considered that the existing isEnum()/assignable checks already gate the *return* value but not the *side effect* of loading? That gap is the whole point here. Added a regression test that resolves a non-enum probe class and asserts its static initializer never ran (fails on the old code, passes now).


garydgregory:
@digi-scrypt Please port to the `1.X` branch.

PR #399, opened 2026-06-09
Reject trailing characters in DecimalLocaleConverter.parse

rootvector2:
`DecimalLocaleConverter.parse` calls `DecimalFormat.parse(String)`, which stops at the first character it cannot read and silently drops the rest, so a locale numeric conversion of `123abc` returns `123` and `42 OR 1=1` returns `42`. Every numeric locale converter (`Byte`/`Short`/`Integer`/`Long`/`Float`/`Double`/`BigDecimal`/`BigInteger`) routes through this one method, while the sibling `DateLocaleConverter` and the non-locale `NumberConverter` already reject leftover input with a `ParsePosition` check. Found by comparing the locale number parsers against `DateLocaleConverter`; switched to that same check so the whole string must be consumed.

- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/locale/converters/DecimalLocaleConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/locale/converters/DecimalLocaleConverter.java
@@ -20,6 +20,7 @@ package org.apache.commons.beanutils2.locale.converters;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.ParseException;
+import java.text.ParsePosition;
 import java.util.Locale;
 
 import org.apache.commons.beanutils2.ConversionException;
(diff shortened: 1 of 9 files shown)


garydgregory:
Hello @rootvector2 
Please review the failing tests. TY!

rootvector2:
Those were the existing `(B)` assertions that pinned `DecimalFormat`'s partial parse of wrong-format input (e.g. `1,234.56` -> `1.234`). Now that the whole string has to be consumed, those inputs fall through to the converter default, so I updated the expected values and the now-stale comments across the numeric locale converter tests. `mvn test` is green.

garydgregory:
Thank you @rootvector2 
Merged 🚀 
Would you mind porting this fix to the `1.X` branch?

rootvector2:
Ported to `1.X` in #400. Same `ParsePosition` check, plus the `(B)` test assertions updated since the wrong-format inputs now fall through to the default.

PR #400, opened 2026-06-21
Reject trailing characters in DecimalLocaleConverter.parse (1.X)

rootvector2:
Port of #399 to the `1.X` branch.

`DecimalLocaleConverter.parse` calls `DecimalFormat.parse(String)`, which stops at the first character it cannot read and silently drops the rest, so a locale numeric conversion of `123abc` returns `123` and `42 OR 1=1` returns `42`. Every numeric locale converter (`Byte`/`Short`/`Integer`/`Long`/`Float`/`Double`/`BigDecimal`/`BigInteger`) routes through this one method, while the sibling `DateLocaleConverter` already rejects leftover input with a `ParsePosition` check. Switched to that same form so the whole string must be consumed, otherwise `ParseException` is thrown.

Added `DecimalLocaleConverterTest` (fails without the runtime change) and updated the now-stale `(B)` assertions across the numeric locale converter tests, which pinned `DecimalFormat`'s old partial-parse values and now fall through to the converter default.

`mvn test -Dtest='*LocaleConverterTest'` is green (104 tests). The only failure in a full `mvn` run is `LocaleBeanificationTest.testContextClassloaderIndependence`, which is a pre-existing flake unrelated to this change (it fails the same way on a clean `1.X` checkout and installs its own mock converters that bypass real parsing).

### DIFF
--- a/src/main/java/org/apache/commons/beanutils/locale/converters/DecimalLocaleConverter.java
+++ b/src/main/java/org/apache/commons/beanutils/locale/converters/DecimalLocaleConverter.java
@@ -20,6 +20,7 @@ package org.apache.commons.beanutils.locale.converters;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.ParseException;
+import java.text.ParsePosition;
 import java.util.Locale;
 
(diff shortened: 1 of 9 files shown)


garydgregory:

PR #401, opened 2026-06-21
Preserve magnitude in BigInteger and BigDecimal locale converters

rootvector2:
`BigIntegerLocaleConverter` and `BigDecimalLocaleConverter` derive their result from the `Long`/`Double` that `DecimalLocaleConverter.parse` returns because `setParseBigDecimal` is never enabled, so a value past `long`/`double` range comes back silently wrong (`9999999999999999999` becomes `9223372036854775807` for `BigInteger` and `1.0E+19` for `BigDecimal`); enabling `setParseBigDecimal` only for those two converters via an `isParseBigDecimal()` hook keeps the parse lossless while the narrowing `Byte`/`Short`/`Integer`/`Long`/`Float`/`Double` converters (and `DoubleLocaleConverter`'s `(Double)` cast) are untouched, found by comparing the locale converters with the non-locale `BigInteger`/`BigDecimal` converters that parse the string directly.

- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/locale/converters/BigDecimalLocaleConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/locale/converters/BigDecimalLocaleConverter.java
@@ -88,4 +88,9 @@ public class BigDecimalLocaleConverter extends DecimalLocaleConverter<BigDecimal
         }
     }
 
+    @Override
+    protected boolean isParseBigDecimal() {
+        return true;
+    }
+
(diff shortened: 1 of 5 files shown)


garydgregory:
Thank you for the PR @rootvector2, merged 🚀 
Would you please port to the `1.X` branch?
TY!

rootvector2:
Done, ported in #403.

PR #404, opened 2026-06-25
Reject out-of-range long and float values in NumberConverter

rootvector2:
`NumberConverter.toNumber(Class, Number)` range-checks the `byte`/`short`/`int` branches but the `Long` branch has no bounds check and the `Float` branch only checks the upper bound, so an out-of-range value (for example a locale-parsed string past `long` range, which `DecimalFormat` returns as a `Double`) is silently clamped to `Long.MAX_VALUE` or to `-Infinity` instead of throwing; found while auditing the numeric converters, fixed by adding the missing bounds checks next to the existing ones.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
@@ -449,6 +449,12 @@ public abstract class NumberConverter<N extends Number> extends AbstractConverte
 
         // Long
         if (targetType.equals(Long.class)) {
+            if (value.doubleValue() > Long.MAX_VALUE) {
+                throw ConversionException.format("%s value '%s' is too large for %s", toString(sourceType), value, toString(targetType));
+            }
+            if (value.doubleValue() < Long.MIN_VALUE) {
+                throw ConversionException.format("%s value '%s' is too small %s", toString(sourceType), value, toString(targetType));
+            }
             return targetType.cast(Long.valueOf(value.longValue()));
         }
 
@@ -456,6 +462,9 @@ public abstract class NumberConverter<N extends Number> extends AbstractConverte
         if (targetType.equals(Float.class)) {
             if (value.doubleValue() > Float.MAX_VALUE) {
                 throw ConversionException.format("%s value '%s' is too large for %s", toString(sourceType), value, toString(targetType));
+            }
+            if (value.doubleValue() < -Float.MAX_VALUE) {
+                throw ConversionException.format("%s value '%s' is too small %s", toString(sourceType), value, toString(targetType));
             }
             return targetType.cast(Float.valueOf(value.floatValue()));
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
```diff
@@ -457,6 +463,9 @@ private <T> T toNumber(final Class<?> sourceType, final Class<T> targetType, fin
             if (value.doubleValue() > Float.MAX_VALUE) {
                 throw ConversionException.format("%s value '%s' is too large for %s", toString(sourceType), value, toString(targetType));
             }
+            if (value.doubleValue() < -Float.MAX_VALUE) {
```
garydgregory, line 466: Why are we not using `MIN_VALUE` here?

src/test/java/org/apache/commons/beanutils2/converters/FloatConverterTest.java
```diff
@@ -70,11 +70,17 @@ void testInvalidAmount() {
         final Converter<Float> converter = makeConverter();
         final Class<?> clazz = Float.class;
         final Double max = Double.valueOf(Float.MAX_VALUE);
+        final Double min = Double.valueOf(-Float.MAX_VALUE);
         final Double tooBig = Double.valueOf(Double.MAX_VALUE);
+        final Double tooSmall = Double.valueOf(-Double.MAX_VALUE);
         // Maximum
         assertEquals(Float.valueOf(Float.MAX_VALUE), converter.convert(clazz, max), "Maximum");
+        // Minimum
+        assertEquals(Float.valueOf(-Float.MAX_VALUE), converter.convert(clazz, min), "Minimum");
```
garydgregory, line 79: Why are we not using MIN_VALUE here?

garydgregory (commented):
Hello @rootvector2 
Please see my 2 comments.

src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
rootvector2: `Float.MIN_VALUE` is the smallest positive value (`1.4E-45`), not the most negative one, so it can't serve as the lower bound. The most negative finite float is `-Float.MAX_VALUE`. The `Long` branch above can use `Long.MIN_VALUE` because for integral types that is the most negative value, but for floating point types it isn't.

src/test/java/org/apache/commons/beanutils2/converters/FloatConverterTest.java
rootvector2: Same reason as in the converter: `Float.MIN_VALUE` is the smallest positive value, not the most negative, so the lowest representable float is `-Float.MAX_VALUE`.

garydgregory:
TY @rootvector2 , merged 🚀 Please verify if this apply to the 1.X branch as well.

rootvector2:
yep, same issue on 1.X. the `Long` branch has no bounds check and the `Float` branch only checks the upper bound there too, so out-of-range input clamps to `Long.MAX_VALUE` / `-Infinity` instead of throwing. ported the fix in #405.

PR #405, opened 2026-06-26
Reject out-of-range long and float values in NumberConverter (1.X)

rootvector2:
Port of #404 to the 1.X branch. @garydgregory asked to verify whether the fix applies there, and it does, identically.

`NumberConverter.toNumber(Class, Number)` range-checks the `byte`/`short`/`int` branches, but the `Long` branch has no bounds check and the `Float` branch only checks the upper bound, so an out-of-range value is silently clamped instead of rejected: a `Double`/`BigInteger`/`BigDecimal` beyond `long` range is truncated/clamped to `Long.MAX_VALUE`, a locale-parsed String past `long` range comes back from `DecimalFormat` as a `Double` and gets clamped the same way, and a `Number` below `-Float.MAX_VALUE` becomes `-Infinity`. Fix adds the missing bounds checks to the `Long` branch and the lower bound to the `Float` branch, mirroring the existing branches, so out-of-range input throws `ConversionException`.

Regression tests added to `LongConverterTest` (`testInvalidAmount`, `testLocaleStringOutOfRange`) and `FloatConverterTest` (negative overflow); they fail without the runtime change. Both converter test classes pass. The full `mvn` run is green except for `LocaleBeanificationTest.testContextClassloaderIndependence`, which already fails on an unmodified 1.X checkout in my environment (it passes in isolation) and is unrelated to this change.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils/converters/NumberConverter.java
+++ b/src/main/java/org/apache/commons/beanutils/converters/NumberConverter.java
@@ -481,6 +481,14 @@ public abstract class NumberConverter extends AbstractConverter {
 
         // Long
         if (targetType.equals(Long.class)) {
+            if (value.doubleValue() > Long.MAX_VALUE) {
+                throw new ConversionException(toString(sourceType) + " value '" + value
+                        + "' is too large for " + toString(targetType));
+            }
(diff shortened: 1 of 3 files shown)


garydgregory:
TY @rootvector2 , merged 🚀

PR #406, opened 2026-06-27
Reject out-of-range values in LongLocaleConverter

rootvector2:
`LongLocaleConverter.parse` narrows the parsed number with `result.longValue()` and never range-checks it, so a value beyond `long` range like `99999999999999999999` is silently clamped to `Long.MAX_VALUE` instead of rejected: `DecimalFormat` returns it as a `Double` (the converter does not set `parseBigDecimal`) and `Double.longValue()` saturates. The sibling narrowing locale converters `IntegerLocaleConverter`, `ByteLocaleConverter`, `ShortLocaleConverter` and `FloatLocaleConverter` already reject out-of-range input, so `LongLocaleConverter` gets the same bounds check before narrowing. Found while auditing the locale converter family against those sibling checks.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/locale/converters/LongLocaleConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/locale/converters/LongLocaleConverter.java
@@ -78,6 +78,11 @@ public class LongLocaleConverter extends DecimalLocaleConverter<Long> {
             return (Long) result;
         }
 
+        final double doubleValue = result.doubleValue();
+        if (doubleValue < Long.MIN_VALUE || doubleValue > Long.MAX_VALUE) {
+            throw new ConversionException("Supplied number is not of type Long: " + result);
+        }
+
         return Long.valueOf(result.longValue());
     }
 }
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @rootvector2 
Please use `assertThrows`. Also port to the `1.X` branch ;)

rootvector2:
Done. Switched the test to `assertThrows` and pushed. 1.X port is up as #407.

garydgregory:
This is unrelated, but could you have a go at fixing the test in Java 25 and up? There are failure in SQL related classes. TY!

rootvector2:
Took a look. The failures are the two `testLocale` cases in `SqlTimeConverterTest` and `SqlTimestampConverterTest`. CLDR (JDK 20+) switched the US SHORT time format to a narrow no-break space (U+202F) before AM/PM, and the tests hard-coded a regular space. Fix is in #410 — it derives the separator from the JVM's own format so it passes on both old and new JDKs. Verified failing/passing on OpenJDK 26, full suite still green on 21.

PR #407, opened 2026-06-27
Reject out-of-range values in LongLocaleConverter (1.X)

rootvector2:
Port of #406 to the `1.X` branch.

`LongLocaleConverter.parse` narrows the parsed number with `result.longValue()` and never range-checks it, so a value beyond `long` range like `99999999999999999999` is silently clamped to `Long.MAX_VALUE` instead of rejected: `DecimalFormat` returns it as a `Double` (the converter does not set `parseBigDecimal`) and `Double.longValue()` saturates.

The sibling narrowing locale converters `IntegerLocaleConverter`, `ByteLocaleConverter`, `ShortLocaleConverter` and `FloatLocaleConverter` already reject out-of-range input, so `LongLocaleConverter` gets the same bounds check before narrowing. Found while auditing the locale converter family against those sibling checks.

Added a regression test to `LongLocaleConverterTest`; it fails without the runtime change (`99999999999999999999` comes back as `9223372036854775807`). `mvn test -Dtest='*LocaleConverterTest'` is green (107 tests).

### DIFF
--- a/src/main/java/org/apache/commons/beanutils/locale/converters/LongLocaleConverter.java
+++ b/src/main/java/org/apache/commons/beanutils/locale/converters/LongLocaleConverter.java
@@ -20,6 +20,8 @@ package org.apache.commons.beanutils.locale.converters;
 import java.text.ParseException;
 import java.util.Locale;
 
+import org.apache.commons.beanutils.ConversionException;
+
 /**
  * Standard {@link org.apache.commons.beanutils.locale.LocaleConverter} implementation that converts an incoming locale-sensitive String into a
  * {@code java.lang.Long} object, optionally using a default value or throwing a {@link org.apache.commons.beanutils.ConversionException} if a conversion error
@@ -172,6 +174,10 @@ public class LongLocaleConverter extends DecimalLocaleConverter {
         final Object result = super.parse(value, pattern);
         if (result == null || result instanceof Long) {
             return result;
+        }
+        final double doubleValue = ((Number) result).doubleValue();
+        if (doubleValue < Long.MIN_VALUE || doubleValue > Long.MAX_VALUE) {
+            throw new ConversionException("Supplied number is not of type Long: " + result);
         }
         return Long.valueOf(((Number) result).longValue());
     }
--- a/src/test/java/org/apache/commons/beanutils/locale/converters/LongLocaleConverterTest.java
+++ b/src/test/java/org/apache/commons/beanutils/locale/converters/LongLocaleConverterTest.java
@@ -17,6 +17,10 @@
 
 package org.apache.commons.beanutils.locale.converters;
 
+import java.text.DecimalFormat;
+
+import org.apache.commons.beanutils.ConversionException;
+
 /**
  * Test Case for the LongLocaleConverter class.
  *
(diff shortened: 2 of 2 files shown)


garydgregory (changes requested):
Hello @rootvector2 
Please use `assertThrows`.

garydgregory:
@rootvector2 ping ;)

rootvector2:
Done, switched the two out-of-range cases to `assertThrows`.

garydgregory:
Merged 🚀 TY! Would you have a look at these warnings please?
```
Jul 02, 2026 3:45:42 PM org.junit.platform.launcher.core.DiscoveryIssueNotifier logIssues
WARNING: TestEngine with ID 'junit-jupiter' encountered 3 non-critical issues during test discovery:

(1) [WARNING] Inner class 'org.apache.commons.collections4.map.AbstractMapTest$TestMapEntrySet' looks like it was intended to be a test class but will not be executed. It must be static or annotated with @Nested.
    Source: ClassSource [className = 'org.apache.commons.collections4.map.AbstractMapTest$TestMapEntrySet', filePosition = null]
            at org.apache.commons.collections4.map.AbstractMapTest$TestMapEntrySet.<no-method>(SourceFile:0)

(2) [WARNING] Inner class 'org.apache.commons.collections4.map.AbstractMapTest$TestMapKeySet' looks like it was intended to be a test class but will not be executed. It must be static or annotated with @Nested.
    Source: ClassSource [className = 'org.apache.commons.collections4.map.AbstractMapTest$TestMapKeySet', filePosition = null]
            at org.apache.commons.collections4.map.AbstractMapTest$TestMapKeySet.<no-method>(SourceFile:0)

(3) [WARNING] Inner class 'org.apache.commons.collections4.map.AbstractMapTest$TestMapValues' looks like it was intended to be a test class but will not be executed. It must be static or annotated with @Nested.
    Source: ClassSource [className = 'org.apache.commons.collections4.map.AbstractMapTest$TestMapValues', filePosition = null]
            at org.apache.commons.collections4.map.AbstractMapTest$TestMapValues.<no-method>(SourceFile:0)
```

TY!

rootvector2:
Traced it. The warnings come from the `commons-collections4` 4.5.0 tests jar, not from beanutils: `BeanMapTest` extends `org.apache.commons.collections4.map.AbstractMapTest`, and in 4.5.0 its `TestMapEntrySet`/`TestMapKeySet`/`TestMapValues` are concrete non-static inner classes left over from the JUnit 3 `BulkTest` mechanism. JUnit platform 5.13+ flags inherited member classes that have test methods but are neither static nor `@Nested`, so they show up whenever `BeanMapTest` is discovered. The warning is accurate, those entrySet/keySet/values suites don't actually run under Jupiter with 4.5.0.

Already fixed on collections master: apache/commons-collections#681 and apache/commons-collections#689 (COLLECTIONS-884) renamed them to `MapEntrySetTest`/`MapKeySetTest`/`MapValuesTest` and made them `@Nested`, so the warnings go away once beanutils picks up collections4 4.6.0. One thing to watch: at that point the view suites will start executing against `BeanMap`, so the bump may need some test adjustments.

Reproduced on `1.X` with `mvn test -Dtest=BeanMapTest`; surefire puts the log in the `*-output.txt` reports rather than the console.

PR #408, opened 2026-07-01
Stop narrowing BigInteger and BigDecimal in StringLocaleConverter

rootvector2:
`StringLocaleConverter.parse` formatted `BigInteger` via `((Number) value).longValue()` and `BigDecimal` via `doubleValue()`, so a value wider than 64 bits was narrowed before formatting and came back silently wrong (`new BigInteger("99999999999999999999")` formatted to `7,766,279,631,452,241,919`); `DecimalFormat.format(Object)` handles `BigInteger`/`BigDecimal` with full magnitude natively, so I format those two types directly and keep the `long`/`double` path for the primitive wrappers. Found while auditing the locale converters against the magnitude handling in `BigIntegerLocaleConverter`/`BigDecimalLocaleConverter`.


garydgregory:
Hi @rootvector2 
Thank you for the PR, merged; please port to the `1.X` branch.

PR #409, opened 2026-07-02
reject non-canonical UUID strings in UUIDConverter

rootvector2:
`UUID.fromString` accepts abbreviated groups like `1-1-1-1-1` and silently canonicalizes them to `00000001-0001-0001-0001-000000000001`, so `UUIDConverter` converts strings that are not the canonical 8-4-4-4-12 form; found while auditing the converters that delegate straight to a lenient JDK parse, and fixed by rejecting any input that does not round-trip to its canonical string (upper case still converts).

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/UUIDConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/UUIDConverter.java
@@ -57,7 +57,14 @@ public final class UUIDConverter extends AbstractConverter<UUID> {
     @Override
     protected <T> T convertToType(final Class<T> type, final Object value) throws Throwable {
         if (UUID.class.equals(type)) {
-            return type.cast(UUID.fromString(String.valueOf(value)));
+            final String stringValue = String.valueOf(value);
+            final UUID uuid = UUID.fromString(stringValue);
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @rootvector2 
-1 unless you can convince us otherwise. This class reflects what Java's own `UUID` class provides through `java.util.UUID.fromString(String)`.
It seems to me that this PR would narrow what the JRE provides out of the box. 

One could argue that `java.util.UUID.fromString(String)` is lenient by accepting input with leading `'0'`s in its compoenents but that's not something we can change.

Where did you get the requirement that the String form of a UUID or a Converter output in general must roundtrip with its String inout? Maybe I missed it.

I've updated the Javadoc which might give readers better expectations.

rootvector2:
Fair enough. I was treating a canonical round-trip as an implicit expectation, but you're right that there's no spec backing it and the converter should mirror `UUID.fromString` rather than narrow what the JRE accepts. The leading-`0` leniency is the JDK's call, not ours. Withdrawing this one, and thanks for the Javadoc clarification.

rootvector2:
Closing per the discussion above.

PR #410, opened 2026-07-02
Fix sql time and timestamp locale tests on java 20+

rootvector2:
Follow-up to the request on #406 to fix the SQL test failures on Java 25 and up.

`SqlTimeConverterTest.testLocale` and `SqlTimestampConverterTest.testLocale` fail on JDK 20+ because CLDR changed the US SHORT time format to use a narrow no-break space (U+202F) before the AM/PM marker instead of a regular space. The locale converter parses using the JVM's own `DateFormat`, so it expects U+202F, while the tests hard-coded a regular space in both the input string and the pattern that computes the expected value (and `toCalendar` parses non-lenient). The non-SQL date test only uses a date pattern, which is why it wasn't affected.

Rather than hard-code either separator, the tests now derive it from the running JVM's SHORT time format, so they pass on both older JDKs (regular space) and JDK 20+ (U+202F).

Verified failing before and passing after on OpenJDK 26; full suite still green on JDK 21.

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This is a test-only fix; there is no runtime change.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/sql/converters/SqlTimeConverterTest.java
+++ b/src/test/java/org/apache/commons/beanutils2/sql/converters/SqlTimeConverterTest.java
@@ -18,7 +18,9 @@
 package org.apache.commons.beanutils2.sql.converters;
 
 import java.sql.Time;
+import java.text.DateFormat;
 import java.util.Calendar;
+import java.util.Date;
 import java.util.Locale;
 
 import org.apache.commons.beanutils2.converters.AbstractDateConverterTest;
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @rootvector2 

Merged 🚀  Please port to the `1.X` branch BUT take into account the copy-pasta clean I did in f5494fd0a0f3b48c9bc140e228bd67571fe1ae47 after merging this PR.

TY!

PR #412, opened 2026-07-07
Suppress class and declaringClass pseudo-properties in BeanMap

rootvector2:
`BeanMap.initialize` introspects the bean with `java.beans.Introspector` directly and registers every property descriptor, so the `class` pseudo-property (`Object.getClass`, present on every bean) and `declaringClass` (`Enum.getDeclaringClass`, on enums) become map entries. `PropertyUtilsBean` registers `SuppressPropertiesBeanIntrospector.SUPPRESS_CLASS`/`SUPPRESS_DECLARING_CLASS` by default to drop exactly these, but `BeanMap` applies no suppression, so `new BeanMap(bean).get("class")` hands back the bean's `Class` and its class loader, and `keySet()`/`entrySet()` enumerate that entry to any caller treating the `BeanMap` as an opaque `Map`. Skip both names in `initialize()`, matching the `"class".equals(name)` guard already in `BeanUtilsBean.copyProperties`.

Found while auditing which introspection paths honor the default `class` suppression; `BeanMap` was the one that reintroduces it. `BeanMapTest.testClassPropertyNotExposed` fails without the runtime change. Updated the sample key/value arrays that pinned the `class` entry.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
+++ b/src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
@@ -26,6 +26,7 @@ import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.time.DayOfWeek;
 import java.util.Map;
 import java.util.function.Function;
 
@@ -295,7 +296,7 @@ class BeanMapTest extends AbstractMapTest<BeanMap, String, Object> {
         assertNull(map.get("class"));
         assertNull(map.getReadMethod("class"));
 
-        final BeanMap enumMap = new BeanMap(java.time.DayOfWeek.MONDAY);
+        final BeanMap enumMap = new BeanMap(DayOfWeek.MONDAY);
         assertFalse(enumMap.containsKey("class"), "class must not be exposed as a property");
         assertFalse(enumMap.containsKey("declaringClass"), "declaringClass must not be exposed as a property");
         assertNull(enumMap.get("declaringClass"));
--- a/src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
+++ b/src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
@@ -17,7 +17,9 @@
 package org.apache.commons.beanutils2;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
```diff
@@ -188,21 +190,20 @@ public Object[] getNewSampleValues() {
...
+                "someIntegerValue", "someStringValue", "someObjectValue", };
         return keys;
     }
 
     // note to self: the sample values were created manually
     @Override
     public Object[] getSampleValues() {
         final Object[] values = { Integer.valueOf(1234), Long.valueOf(1298341928234L), Double.valueOf(123423.34), Float.valueOf(1213332.12f),
-                Short.valueOf((short) 134), Byte.valueOf((byte) 10), Character.valueOf('a'), Integer.valueOf(1432), "SomeStringValue", objectInFullMap,
-                BeanWithProperties.class, };
```
garydgregory, line 205: Why was `BeanWithProperties` removed here?

src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
```diff
@@ -283,6 +284,23 @@ void testBeanMapClone() {
...
+     * matching the suppression {@link PropertyUtilsBean} applies by default.
+     */
+    @Test
+    void testClassPropertyNotExposed() {
+        final BeanMap map = new BeanMap(new BeanWithProperties());
+        assertFalse(map.containsKey("class"), "class must not be exposed as a property");
+        assertNull(map.get("class"));
+        assertNull(map.getReadMethod("class"));
+
+        final BeanMap enumMap = new BeanMap(java.time.DayOfWeek.MONDAY);
```
garydgregory: Don't use a fully-qualified class name here.

garydgregory (changes requested):
Hi @rootvector2 
Please my see my comments.

src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
rootvector2: `getSampleValues()` is index-aligned with `getSampleKeys()` in `AbstractMapTest`; `BeanWithProperties.class` was the sample value for the `class` key, so removing the key means removing its value too, otherwise the arrays go out of sync.

src/test/java/org/apache/commons/beanutils2/BeanMapTest.java
rootvector2: done, switched to an import of `DayOfWeek`.

PR #414, opened 2026-07-12
Make MethodUtils CACHE_ENABLED volatile

rootvector2:
`setCacheMethods` writes the `@GuardedBy`-annotated `CACHE_ENABLED` flag under the class monitor but `computeIfAbsent` reads it lock-free on the method-lookup path, so a runtime `setCacheMethods(false)` may never be observed by other threads and they keep populating and reading the shared static `CACHE`; marking the field `volatile` restores cross-thread visibility while writes stay synchronized so the flag flip and `CACHE.clear()` remain atomic. Found while auditing the `@GuardedBy` field; a memory-visibility fix has no deterministic unit test, so none is added.

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/MethodUtils.java
+++ b/src/main/java/org/apache/commons/beanutils2/MethodUtils.java
@@ -118,7 +118,7 @@ public final class MethodUtils {
      * would mean having a map keyed by context classloader which may introduce memory-leak problems.
      * </p>
      */
-    private static boolean CACHE_ENABLED = true; //NOPMD @GuardedBy("this")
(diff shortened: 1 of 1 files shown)


garydgregory:
Thank you @rootvector2 , merged 🚀 , please port to 1.x.

rootvector2:
thanks for merging. 1.x port is up in #416, the field is `CACHE_METHODS` there but it's the same one-line fix.

PR #415, opened 2026-07-13
Case-fold column and bean property names with Locale.ROOT

rootvector2:
`AbstractJdbcDynaClass.createDynaProperty` folds JDBC column names to lower case with `String.toLowerCase()`, and `DefaultBeanIntrospector.handleIndexedPropertyDescriptors` upper-cases the first letter of an indexed property name with `String.toUpperCase()`, both using the default locale. Under a Turkish locale `"ID".toLowerCase()` is `"ıd"` (dotless i), so a column `ID` maps to property `ıd` and `getDynaProperty("id")` returns `null`; the introspector likewise rebuilds an indexed accessor as `getİmages` instead of `getImages`. Found from the two `DM_CONVERT_CASE` SpotBugs findings on this repo. Both now fold with `Locale.ROOT`, matching `MappedPropertyDescriptor` (`Character.toUpperCase`) and the converters (`AbstractConverter.toLowerCase` via `StringUtils.toRootLowerCase`) which already fold case locale-independently. Added a `ResultSetDynaClass` regression test that fails under the Turkish locale without the runtime change.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/DefaultBeanIntrospector.java
+++ b/src/main/java/org/apache/commons/beanutils2/DefaultBeanIntrospector.java
@@ -23,6 +23,7 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Method;
 import java.util.List;
+import java.util.Locale;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


garydgregory (changes requested):
@rootvector2 
Thank you for the PR. Use `@DefaultLocale` from JUnit Pioneer.

garydgregory:
Thank you @rootvector2 , merged 🚀 Please port to 1.x..

rootvector2:
thanks for merging. port to 1.X is up in #417, with the test switched to `@DefaultLocale` from junit-pioneer per your review (added 1.9.1 as a test dep since 2.x needs Java 11). can switch the master test over to `@DefaultLocale` in a follow-up too if you want it there.

PR #418, opened 2026-07-14
Make WrapDynaClass instance cache thread-safe

rootvector2:
`WrapDynaClass.createDynaClass` populates its per-classloader cache with `computeIfAbsent`, but the cache is a plain `WeakHashMap`, so threads that share a context classloader race on an unsynchronized map where `computeIfAbsent` is not atomic: concurrent callers can build and return different `WrapDynaClass` instances for one key, and the structural writes can corrupt the map. Switch the cache to `BeanUtils.createCache()`, the weak-keyed `ConcurrentReferenceHashMap` that `MethodUtils`, `ConvertUtilsBean` and `PropertyUtilsBean` already use. Found while auditing the package's `computeIfAbsent` call sites against those thread-safe caches. Like the `MethodUtils` visibility fix this race has no deterministic unit test so none is added, though a stress loop over concurrent `createDynaClass` calls does surface the duplicate instances.

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/WrapDynaClass.java
+++ b/src/main/java/org/apache/commons/beanutils2/WrapDynaClass.java
@@ -23,7 +23,6 @@ import java.lang.ref.SoftReference;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.WeakHashMap;
 
 /**
  * Implements {@link DynaClass} to wrap standard JavaBean instances.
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
@rootvector2 
Please add a unit test to avoid a future regression.

rootvector2:
added a regression test that races 16 threads on `createDynaClass` for one bean class over 200 rounds and asserts a single cached instance comes back. it fails against the old `WeakHashMap` and passes with `BeanUtils.createCache()`.

garydgregory:
Merged, TY @rootvector2 , please port to 1.X if possible.

rootvector2:
thanks for merging. port is up in #419 — `1.X` has no atomic `computeIfAbsent` to lean on, so it swaps the cache to `BeanUtils.createCache()` and synchronizes the get/create/put in `createDynaClass`, with the same regression test ported.

PR #420, opened 2026-07-16
Reject out-of-range color components in ColorConverter

rootvector2:
`ColorConverter.parseToStringColor` already rejects color components above 255, but the `JAVA_COLOR_PATTERN` label prefix `[A-Za-z\d._]+` includes `\d`, so for a bare triple whose first component has four or more digits the greedy label swallows the leading digit and the first capture group only sees the trailing one to three digits. `1234,5,6` is parsed as `Color(4,5,6)`, and `1000,0,0` / `2550,0,0` as black, so a malformed out-of-range value is accepted as a different in-range color instead of ever reaching the `> 255` check. Binding the optional class-name label to its opening `[` (`(?:[A-Za-z\d._]*\[)?`) stops it absorbing digits from the numeric groups; every string `Color#toString()` emits still parses the same, including class names that contain digits. Found while auditing the converters that parse untrusted strings.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/ColorConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/ColorConverter.java
@@ -39,7 +39,7 @@ public class ColorConverter extends AbstractConverter<Color> {
     private static final String HEX_COLOR_PREFIX = "#";
 
     /** Regular expression matching the output of {@link Color#toString()}. */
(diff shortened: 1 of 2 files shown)


garydgregory:
TY @rootvector2 , merged 🚀 Please port to 1.X with the additional assertions in the new test.

rootvector2:
checked the 1.X branch and it has no `ColorConverter` (no `java.awt.Color` handling at all, the converter only came in with 2.x via #47), so there's nothing to port there. thanks for merging.

PR #421, opened 2026-07-17
Pass configured locale to SimpleDateFormat in DateTimeConverter

rootvector2:
`getFormat(String)` builds the pattern `SimpleDateFormat` with the jvm default locale and ignores the converter's configured `locale`, so a converter given both `setLocale` and a pattern parses and formats month and day names in the default locale instead of the configured one; pass `locale` when it is set, matching `DateLocaleConverter` and `StringLocaleConverter`. found auditing locale handling across the date converters.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/converters/AbstractDateConverterTest.java
+++ b/src/test/java/org/apache/commons/beanutils2/converters/AbstractDateConverterTest.java
@@ -269,6 +269,27 @@ public abstract class AbstractDateConverterTest<T> {
     }
 
     /**
+     * Test that a configured Locale is honored when a pattern is also set.
+     */
+    @Test
+    void testLocaleWithPattern() {
+        // Pin the default Locale to one whose month names differ from the configured Locale.
+        final Locale defaultLocale = Locale.getDefault();
+        Locale.setDefault(Locale.US);
+        try {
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/beanutils2/converters/AbstractDateConverterTest.java
```diff
@@ -268,6 +268,27 @@ public void testLocale() {
         Locale.setDefault(defaultLocale);
     }
 
+    /**
+     * Test that a configured Locale is honored when a pattern is also set.
+     */
+    @Test
```
garydgregory, line 275: Use JUnit Pioneer to set a temporary Locale.

garydgregory (changes requested):
Hi @rootvector2 Please see my comment.

src/test/java/org/apache/commons/beanutils2/converters/AbstractDateConverterTest.java
rootvector2: Done, switched the test to Pioneer's `@DefaultLocale` and dropped the manual save/restore. Added `junit-pioneer` as a test dependency since it wasn't in the pom yet (version comes from commons-parent). All 9 subclass runs still pass.

garydgregory:
TY @rootvector2 Merged 🚀 Please port to 1.x if appropriate.

PR #422, opened 2026-07-19
Fix ClassCastException in DoubleLocaleConverter when input is non-Double Number

rootvector2:
`DoubleLocaleConverter.parse()` only handled the `Long` case for non-String inputs; any other `Number` subtype (e.g. `Integer`, `Float`, `BigDecimal`) passed directly to the converter was unchecked-cast to `Double`, throwing `ClassCastException` instead of a proper conversion.


garydgregory:
Merged 🚀 Thank you @rootvector2 , please port to 1.x.

rootvector2:
thanks for merging, will get the 1.x port up.

rootvector2:
thanks for the merge. 1.x port is up in #424. same root cause there, though it surfaces as a `ConversionException` instead since the cast happens in `BaseLocaleConverter` rather than in `parse`.

PR #423, opened 2026-07-20
Reject trailing separators in PointConverter

rootvector2:
`PointConverter.convertToType` splits the inner coordinate string with `POINT_SPLIT.split(coordinates)`, and `String.split` with the default limit drops trailing empty tokens. A trailing separator after the two coordinates is swallowed, so `(100,200,)` and `(100,200,,)` parse to `Point(100,200)`. That is inconsistent with `(100,)`, which is already rejected because it yields a single token.

Pass `-1` as the split limit so trailing empties are kept and the existing `xy.length != 2` check rejects the malformed input. Keeping the bound in the converter means every caller gets the same validation, and valid `(x,y)` forms are unchanged. Found while auditing the awt converters alongside `ColorConverter`.


garydgregory:
@rootvector2 
Please see https://github.com/apache/commons-beanutils/pull/419#issuecomment-5036216965

PR #426, opened 2026-07-22
Resolve column name in ResultSetIterator.set

rootvector2:
`set` passes the raw lower-cased property name to `ResultSet.updateObject` while the read path resolves it through `getColumnName` first, so under the default `lowerCase` an update to a mixed-case column targets the wrong column or throws on a case-sensitive driver; found diffing the get and set paths against `getObject`.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils2/sql/DynaResultSetTest.java
+++ b/src/test/java/org/apache/commons/beanutils2/sql/DynaResultSetTest.java
@@ -25,6 +25,8 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.math.BigDecimal;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.util.Iterator;
 
 import org.apache.commons.beanutils2.DynaBean;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/beanutils2/sql/DynaResultSetTest.java
```diff
@@ -96,6 +98,26 @@ void testGetName() {
...
     }
 
+    /**
+     * With the default {@code lowerCase} option the property name differs from the real column name, and the read path resolves it through
+     * {@code getColumnName}. Verify that {@code set} resolves it the same way, so the update targets the real column name and not the lower-cased property
+     * name.
+     */
+    @Test
+    void testSetUsesColumnName() throws Exception {
+        final String[] updatedColumn = new String[1];
```
garydgregory: @rootvector2 
Use an `AtomicReference` instead of the old-school array hack ;)

garydgregory (changes requested):
Hi @rootvector2 
Please see my comment.

garydgregory:
@rootvector2 Ping 🔔

src/test/java/org/apache/commons/beanutils2/sql/DynaResultSetTest.java
rootvector2: done, switched to `AtomicReference`.

garydgregory:
@rootvector2 
TY, merged 🚀 Please port to 1.x.

rootvector2:
thanks for merging this. 1.x port is up as #428.

PR #428, opened 2026-07-24
Resolve column name in ResultSetIterator.set (1.X)

rootvector2:
Port of #426 to 1.X. `set` passes the raw lower-cased property name to `ResultSet.updateObject` while the read path resolves it through `getColumnName` first, so under the default `lowerCase` an update to a mixed-case column targets the wrong column or throws on a case-sensitive driver.

Same one-line fix and regression test as master, with the test adapted to the JUnit 3 style on this branch (and using `AtomicReference` per the review on #426). `DynaResultSetTest.testSetUsesColumnName` fails on clean `1.X` with `expected:<string[P]roperty> but was:<string[p]roperty>` and passes with the change.

Note: as on #417, `mvn` on a clean `1.X` checkout fails `LocaleBeanificationTest#testContextClassloaderIndependence` in my environment (it passes in isolation), so no fully green default-goal run locally; the failure is identical with and without this change. `apache-rat:check` and `checkstyle:check` pass.

### DIFF
--- a/src/test/java/org/apache/commons/beanutils/DynaResultSetTest.java
+++ b/src/test/java/org/apache/commons/beanutils/DynaResultSetTest.java
@@ -18,7 +18,10 @@
 package org.apache.commons.beanutils;
 
 import java.math.BigDecimal;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicReference;
 
 import junit.framework.TestCase;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/beanutils/DynaResultSetTest.java
```diff
@@ -123,6 +126,28 @@ public void testGetName() {
...
     }
 
+    /**
+     * With the default {@code lowerCase} option the property name differs from the real column name, and the read path resolves it through
+     * {@code getColumnName}. Verify that {@code set} resolves it the same way, so the update targets the real column name and not the lower-cased property
+     * name.
+     */
+    public void testSetUsesColumnName() throws Exception {
+
+        final AtomicReference<String> updatedColumn = new AtomicReference<String>();
```
garydgregory, line 136: Hi @rootvector2 
In the future, you do not need the extra generics noise, you can simply say:
```java
final AtomicReference<String> updatedColumn = new AtomicReference<>();
```

rootvector2: makes sense, I'll use the diamond going forward. thanks for merging this.

PR #432, opened 2026-08-01
Reject negative index values in DefaultResolver.getIndex

rootvector2:
`getIndex` parses the subscript with `Integer.parseInt` and returns it unvalidated, but the same method uses `-1` as its documented "not indexed" sentinel, so `getIndex("stringArray[-1]")` is indistinguishable from `getIndex("stringArray")` and `getIndex("stringArray[-7]")` hands the caller a negative subscript. `PropertyUtilsBean.getIndexedProperty`/`setIndexedProperty` are unaffected because they have their own `index < 0` guard, but the three callers that branch on `index >= 0` fall through to the non-indexed path and write the whole property instead of one element. `BeanUtils.copyProperty(bean, "stringArray[-1]", "x")` and `LocaleBeanUtils.setProperty(bean, "stringArray[-1]", "x")` both replace a five-element `String[]` with `["x"]` and throw nothing, while `BeanUtils.setProperty` and `PropertyUtils.setIndexedProperty` reject the same expression. Found while auditing `DefaultResolver` against its call sites after noticing `getIndex` can return a value that collides with its own sentinel.

Indexes are zero-relative, so a negative subscript is rejected in `getIndex` with the same `IllegalArgumentException` it already throws for a non-numeric one. Keeping the check where the subscript is parsed covers all five call sites at once, and `DefaultResolver` is the only `Resolver` implementation. No existing test used a negative index; the new assertions in `DefaultResolverTest.testGetIndex` fail without the runtime change.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/expression/DefaultResolver.java
+++ b/src/main/java/org/apache/commons/beanutils2/expression/DefaultResolver.java
@@ -57,7 +57,7 @@ public class DefaultResolver implements Resolver {
      *
      * @param expression The property expression
      * @return The index value or -1 if the property is not indexed
-     * @throws IllegalArgumentException If the indexed property is illegally formed or has an invalid (non-numeric) value.
+     * @throws IllegalArgumentException If the indexed property is illegally formed or has an invalid (non-numeric or negative) value.
      */
     @Override
     public int getIndex(final String expression) {
@@ -82,6 +82,9 @@ public class DefaultResolver implements Resolver {
                 try {
                     index = Integer.parseInt(value, 10);
                 } catch (final Exception e) {
+                    throw new IllegalArgumentException("Invalid index value '" + value + "'");
+                }
+                if (index < 0) {
                     throw new IllegalArgumentException("Invalid index value '" + value + "'");
                 }
                 return index;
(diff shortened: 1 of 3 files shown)


garydgregory (commented):
@rootvector2 
Java has am unchecked `IndexOutOfBoundsException`. Should it be used here? Are there other call sites that use `IAE` for this category of problems instead of `IndexOutOfBoundsException`?

rootvector2:
The repo's convention is `IAE` for this category: `PropertyUtilsBean.getIndexedProperty` and `setIndexedProperty` both throw `IllegalArgumentException` when the index parsed from the expression is negative ("Invalid indexed property ..."), and `getIndex` itself already throws `IAE` for a non-numeric subscript with the same "Invalid index value" message this check reuses. `IndexOutOfBoundsException` is used for the other category, a well-formed index that falls outside the actual array or `List` at access time; `PropertyUtilsBean` propagates it from the underlying access. A negative subscript is a malformed expression rather than an out-of-range access, and at parse time there's no collection to be out of bounds of. One more practical point: both `PropertyUtilsBean` call sites catch `IllegalArgumentException` from `resolver.getIndex` and rewrap it with bean-class context, so an `IndexOutOfBoundsException` thrown there would bypass those handlers and change their existing message contract.

PR #434, opened 2026-08-11
Record element type of first element in LazyDynaList.transform

rootvector2:
`transform` never records the element type for the first non-null element, so the homogeneity check at the end of the method is dead code and mismatched types are silently accepted (`add(new HashMap<>())` then `add("a POJO String")` is accepted), and `toArray()`/`toDynaBeanArray()` throw `NullPointerException` from `Array.newInstance(elementType, size())` because `elementType` stays null; found while auditing the lazy dyna utilities, fixed by recording the type from the first populated element, mirroring `setElementDynaClass`, which also restores the documented `Map[]` round-trip in Example 1.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/LazyDynaList.java
+++ b/src/main/java/org/apache/commons/beanutils2/LazyDynaList.java
@@ -659,7 +659,18 @@ public class LazyDynaList extends ArrayList<Object> {
 
         // Check the new element type, matches all the
         // other elements in the List
-        if (elementType != null && !newElementType.equals(elementType)) {
+        if (elementType == null) {
+            // The first element populated defines the element type (see class
+            // Javadoc). Record it so later elements are type-checked and
+            // toArray()/toDynaBeanArray() know the element type.
+            this.elementType = newElementType;
+            this.elementDynaBeanType = newDynaBeanType;
+            if (WrapDynaBean.class.isAssignableFrom(newDynaBeanType)) {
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @rootvector2 

Thank you for the PR. What about adding tests for:

- POJO first element -> WrapDynaBean path, toArray() returning the POJO class and toDynaBeanArray() returning WrapDynaBean[].
- DynaBean first element -> elementType = DynaBean subclass, elementDynaBeanType = same subclass.
- addAll(Collection) and addAll(int,Collection) with an untyped list.
- get/growList after type has been fixed by first population.
- toDynaBeanArray() type correctness for the untyped case.

TY!

rootvector2:
Added all five: POJO first element (`toArray()` returns `TestBean[]`, `toDynaBeanArray()` returns `WrapDynaBean[]`), `BasicDynaBean` first element (both arrays typed to the subclass), `addAll(Collection)` and `addAll(int, Collection)` on an untyped list, `get()` growing after the first element fixes the type, and `toDynaBeanArray()` component type for the untyped map case. `LazyDynaListTest` is up to 21 tests, full `mvn` default goal green.

garydgregory:
@rootvector2 Please rebase on git master to resolve a test snafu.

rootvector2:
Rebased on current master, full `mvn` default goal is green here.

garydgregory:
@rootvector2 Thank you, merged 🚀

PR #435, opened 2026-08-14
Clear descriptor caches when bean introspectors change

rootvector2:
`addBeanIntrospector`, `removeBeanIntrospector` and `resetBeanIntrospectors` mutate the `introspectors` list but never invalidate `descriptorsCache`, which `getIntrospectionData` fills once per bean class, so registering a `SuppressPropertiesBeanIntrospector` (the documented BEANUTILS-463 hardening) after a class has already been introspected is a no-op: `getPropertyDescriptor` returns the stale cached descriptor at its early return before it reaches the `isPropertySuppressed` guard, and the suppressed property stays readable and writable via `populate`/`getProperty`/`setProperty`. found while auditing the suppression path from #413/#431, whose own tests call `clearDescriptors()` right before `addBeanIntrospector` to work around this. clearing the descriptor caches when the introspector set changes makes the change take effect for already-cached classes; i did not reuse `clearDescriptors()` because it also flushes the JVM-global `Introspector` cache.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java
+++ b/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java
@@ -136,6 +136,7 @@ public class PropertyUtilsBean {
      */
     public void addBeanIntrospector(final BeanIntrospector introspector) {
         introspectors.add(Objects.requireNonNull(introspector, "introspector"));
+        clearDescriptorCaches();
     }
 
     /**
@@ -146,6 +147,15 @@ public class PropertyUtilsBean {
         descriptorsCache.clear();
         mappedDescriptorsCache.clear();
         Introspector.flushCaches();
+    }
+
+    /**
+     * Discards the memoized introspection results after the registered {@link BeanIntrospector} set has changed. Unlike {@link #clearDescriptors()} this leaves
+     * the JVM-global {@link Introspector} cache untouched.
+     */
+    private void clearDescriptorCaches() {
+        descriptorsCache.clear();
+        mappedDescriptorsCache.clear();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @rootvector2 

Thank you for the PR.

What do you think about adding tests that cover the following patterns:

- `resetBeanIntrospectors()` cache invalidation. The method now calls `clearDescriptorCaches()`, but there is no dedicated test that warms a cache, calls `resetBeanIntrospectors()` ,and checks that a previously suppressed or modified property is re-evaluated.
- Invalidation of the mappedDescriptorsCache path specifically. The existing `testCustomIntrospectionSuppressedMappedProperty*` tests exercise suppression of mapped properties, but they add the introspector before the first access. A test that warms a mapped descriptor, then adds and/or removes an introspector and checks the mapped descriptor is re-created would close the gap.
- Multiple sequential adds/removes.

rootvector2:
makes sense, added all three in 85eda71: `testResetBeanIntrospectorsInvalidatesCache` warms the cache with a suppressor and checks the property is re-evaluated after the reset; `testIntrospectorChangeInvalidatesMappedDescriptorCache` warms a mapped descriptor, then adds/removes an introspector and asserts with `assertNotSame` that the cached `MappedPropertyDescriptor` is re-created rather than served stale (the suppression guard alone would mask a stale entry, so the identity check is what closes that gap); `testSequentialIntrospectorChangesInvalidateCache` does two adds then two removes and checks the visible property set after each step. verified all three fail with the invalidation in `PropertyUtilsBean` reverted. full default `mvn` goal green, 1291 tests.

garydgregory:
@rootvector2 Please rebase on git master to pick up a fix to the tests.

rootvector2:
rebased on master, full `mvn` run green with 1291 tests.

garydgregory:
@rootvector2 
Thank you, merged 🚀

PR #436, opened 2026-08-16
Reject out-of-range sql date/time fields in DateTimeConverter.toDate

rootvector2:
`DateTimeConverter.toDate` builds `java.sql.Date`/`Time`/`Timestamp` with `valueOf`, which silently rolls over out-of-range fields (`2006-02-31` -> `2006-03-03`, `25:70:90` -> `02:11:30`, `2006-10-23 25:70:90` -> next day), so malformed input in the default `useLocaleFormat=false` path is accepted as a different value instead of rejected; found while auditing the sql converters against the strict `setLenient(false)`/`ParsePosition` check the same class already applies on its locale-format branch, and fixed by validating against strict JDBC-format `DateTimeFormatter`s (`ResolverStyle.STRICT`) before `valueOf`, leaving valid padded and single-digit input unchanged.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/DateTimeConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/DateTimeConverter.java
@@ -22,10 +22,16 @@ import java.text.SimpleDateFormat;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
+import java.time.format.SignStyle;
+import java.time.temporal.ChronoField;
 import java.time.temporal.TemporalAccessor;
 import java.util.Calendar;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


garydgregory:
Hello @rootvector2 

Thank you for the PR.

The Javadoc  and exception messages should be updated, for example

> String must be in JDBC format [yyyy-MM-dd] to create a ...

The behavioral contract has now become stricter: out-of-range fields are rejected instead of being rolled over. The error messages are unchanged. The Javadoc and messages are technically still true but it do not document the new
strictness. How about mention that "validation is strict, out-of-range fields are rejected”.

rootvector2:
Makes sense. Updated all three exception messages to append "validation is strict, out-of-range fields are rejected" and added the same note to the `toDate` javadoc. Full `mvn` build still green (1287 tests, 0 failures).

garydgregory:
Hello @rootvector2 

I pushed patched up version of this PR. I think this PR can be closed unless you see follow up work needed.

The problematic bug I fixed is three instances of double parsing.

Instead of going round and round, I applied the diff locally, made the following changes, and pushed to master.

Double parsing for the `java.sql.Date` case.

Validation parses with the strict formatter and valueOf parses again. Functionally correct, but wasteful. For example:
```
LocalDate.parse(value, SQL_DATE_FORMAT);
return type.cast(java.sql.Date.valueOf(value));
```
The parsed LocalDate is discarded. A more efficient form would be `Date.valueOf(LocalDate.parse(...))`.

Double parsing of for the `java.sql.Time` case. Same as above.

Double parsing of for the `java.sql.Timestamp` case. Same as above.

I added details to Javadocs to clarify intent.

I added the follow unit tests:

In SqlDateConverterTest

testDefaultStringToTypeConvertValidLeapYear
testDefaultStringToTypeConvertInvalidNonLeapYearFeb29
testDefaultStringToTypeConvertValidBoundaryDates
testDefaultStringToTypeConvertStrictValidationMessage

In SqlTimeConverterTest

testDefaultStringToTypeConvertValidBoundaryTimes
testDefaultStringToTypeConvertInvalidTimeHourMinuteSecond
testDefaultStringToTypeConvertStrictValidationMessage

In SqlTimestampConverterTest

testDefaultStringToTypeConvertValidTimestampWithFraction
testDefaultStringToTypeConvertInvalidTimestampOutOfRangeDate
testDefaultStringToTypeConvertInvalidTimestampOutOfRangeTime
testDefaultStringToTypeConvertStrictValidationMessage

These tests cover valid leap-year and boundary values, invalid out-of-range values for all three SQL types, timestamp fraction handling and verification that the new ConversionException messages mention strict validation.

You are credited in `changes.xml`.

rootvector2:
Confirmed 148cad5 on master covers everything, no follow up needed. Agreed on `valueOf(LocalDate.parse(...))`, cleaner than parsing twice. Thanks for landing it and the credit.

PR #437, opened 2026-08-22
Skip static methods in FluentPropertyBeanIntrospector and mapped lookup

rootvector2:
`FluentPropertyBeanIntrospector.introspect` turns every public method whose name starts with the write prefix into a write method without checking `Modifier.isStatic`, and `MappedPropertyDescriptor.getMethod` accepts the static methods `getMatchingAccessibleMethod` returns (unlike its sibling `internalGetMethod`, which already skips them), so once the fluent introspector is registered `BeanUtils.setProperty(form, "locale.default", "fr-FR")` on any bean with a `Locale` getter invokes the static `Locale.setDefault` and any public static `setX(T)` or `getX(String)`/`setX(String, T)` reachable through a nested property becomes writable from a `populate` map; found while auditing the introspectors against the JavaBeans rule `java.beans.Introspector` follows (static methods are never accessors), fixed by skipping static methods at both lookup sites, with regression tests in `FluentPropertyBeanIntrospectorTest` and `MappedPropertyTest` that fail without the runtime change.


garydgregory:
@rootvector2 merged 🚀 Thank you.

PR #438, opened 2026-09-01
NumberConverter.toNumber(Class, Number): Preserve float and double magnitude when converting to BigInteger

rootvector2:
`NumberConverter.toNumber(Class, Number)` converts a `Float` or `Double` source to `BigInteger` through `BigInteger.valueOf(value.longValue())`, so a floating-point magnitude past `long` range silently saturates instead of keeping its value: `converter.convert(BigInteger.class, 1e30d)` returns `9223372036854775807` and `1e20f` returns the same, not the real magnitude.

the sibling `BigDecimal` branch just above already formats `Float`/`Double` with `new BigDecimal(value.toString())` to avoid exactly this, and #401 / #408 established that `BigInteger` and `BigDecimal` conversions must preserve magnitude. this routes floating-point sources through `new BigDecimal(value.toString()).toBigInteger()`; integer sources keep the exact `longValue()` path and in-range values are unchanged. found while auditing the `toNumber` narrowing family after #404, which only added the `Long` and `Float` range checks and never touched the `BigInteger` branch.

`BigIntegerConverterTest.testLargeFloatingPointMagnitude` fails before the change (`9223372036854775807`) and passes after.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
@@ -493,6 +493,9 @@ public abstract class NumberConverter<N extends Number> extends AbstractConverte
             if (value instanceof BigDecimal) {
                 return targetType.cast(((BigDecimal) value).toBigInteger());
             }
+            if (value instanceof Float || value instanceof Double) {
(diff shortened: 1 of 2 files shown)


garydgregory:

PR #439, opened 2026-09-05
fix range checks bypassed by wraparound in number converters

rootvector2:
the range checks in `NumberConverter.toNumber(Class, Number)` and in the byte/short/integer/long locale converters run on `longValue()`/`doubleValue()` of the source, but `longValue()` of a `BigInteger` or `BigDecimal` keeps only the low-order 64 bits and a `double` cannot represent 2^63 - 1, so out-of-range values wrap or round into range and convert to unrelated results (`BigInteger` 2^63 becomes `Long.MIN_VALUE`, 2^64 + 5 becomes `Integer` 5, and a locale-parsed `"9223372036854775808"` is clamped to `Long.MAX_VALUE`); found while checking the bounds from #404 against boundary values, fixed by comparing the exact value as `BigDecimal` before narrowing.

- [x] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) if you use Artificial Intelligence (AI).
- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?
  - Claude (Anthropic) was used to find the bug and to write the patch, the tests and this description. This account is responsible for what it submits.

### DIFF
--- a/src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
+++ b/src/main/java/org/apache/commons/beanutils2/converters/NumberConverter.java
@@ -318,6 +318,59 @@ public abstract class NumberConverter<N extends Number> extends AbstractConverte
     }
 
     /**
+     * Converts a {@code Number} to a {@code long}, validating that its whole part is within the specified range.
+     * <p>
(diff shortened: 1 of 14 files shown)


rootvector2:
Ticked the AI-use boxes in the description. Any update?

garydgregory:
This PR is just one of many in my queue to process...

